PR #53, opened 2023-02-14
Tonic retry client

george-james-mo:
1. error handling:
the client fails immediately if x-token is invalid
the client fails immediately if URI is invalid
2. define error type according to https://github.com/rpcpool/solana-geyser-grpc/issues/43
3. check x_token length == 28, https://github.com/rpcpool/solana-geyser-grpc/issues/45

### DIFF
--- a/src/bin/client.rs
+++ b/src/bin/client.rs
@@ -329,14 +329,19 @@ mod tests {
     async fn test_channel_invalid_token_none() {
         let endpoint = "http://127.0.0.1:10000".to_owned();
         let x_token = None;
-        assert!(matches!(RetryChannel::new(endpoint, x_token), Err(Error::XToken(_))));
+        assert!(matches!(
+            RetryChannel::new(endpoint, x_token),
+            Err(Error::XToken(_))
+        ));
     }
 
     #[tokio::test]
     async fn test_channel_invalid_uri() {
         let endpoint = "sites/files/images/picture.png".to_owned();
         let x_token = "1234567891012141618202224268".to_owned();
-        assert!(matches!(RetryChannel::new(endpoint, Some(x_token)), Err(Error::InvalidUri(_))));
-        dbg!(&res);
+        assert!(matches!(
+            RetryChannel::new(endpoint, Some(x_token)),
+            Err(Error::InvalidUri(_))
+        ));
     }
 }


src/bin/client.rs
```diff
@@ -107,51 +261,93 @@ async fn main() -> anyhow::Result<()> {
...
+    async fn test_channel_invalid_token_none() {
+        let endpoint = "http://127.0.0.1:10000".to_owned();
+        let x_token = None;
+        let res: Result<RetryChannel, Error> = RetryChannel::new(endpoint, x_token);
+
+        if let Err(Error::XToken(_)) = res {
+            assert!(true);
+        } else {
+            assert!(false);
         }
```
nbaksalyar: I think this can be simplified as
```suggestion
        assert!(matches!(RetryChannel::new(endpoint, x_token), Err(Error::XToken(_))));
```

src/bin/client.rs
```diff
@@ -107,51 +261,93 @@ async fn main() -> anyhow::Result<()> {
...
+    #[tokio::test]
+    async fn test_channel_invalid_uri() {
+        let endpoint = "sites/files/images/picture.png".to_owned();
+        let x_token = "1234567891012141618202224268".to_owned();
+        let res: Result<RetryChannel, Error> = RetryChannel::new(endpoint, Some(x_token));
+        if let Err(Error::InvalidUri(_)) = res {
+            assert!(true);
+        } else {
+            assert!(false);
+        }
```
nbaksalyar: Same here:
```suggestion
        assert!(matches!(RetryChannel::new(endpoint, Some(x_token)), Err(Error::InvalidUri(_))));
```

src/bin/client.rs
```diff
@@ -107,51 +261,93 @@ async fn main() -> anyhow::Result<()> {
...
 
-    let is_https = args.endpoint.parse::<Uri>()?.scheme_str() == Some("https");
-    let mut endpoint = Endpoint::from_shared(args.endpoint.clone())?;
-    if is_https {
-        endpoint = endpoint.tls_config(ClientTlsConfig::new())?;
+    // Client with retry policy
+    let res: Result<RetryChannel, Error> =
+        RetryChannel::new(args.endpoint.clone(), args.x_token.clone());
+    if let Err(e) = res {
+        println!("Error: {}", e);
```
nbaksalyar: Minor thing, but I'd suggest using `eprintln!` for error cases:
```suggestion
        eprintln!("Error: {}", e);
```
this would simplify using the client as a *nix-style program - e.g., parse logs in `stdout` and `stderr` separately.

PR #62, opened 2023-02-22
chore: mv client.rs into rust dir

george-james-mo:
i think the best practice is:
a Cargo.toml inlcude just [workspace]
libsolana_geyser_grpc in 'cargo new lib-geyser-grpc'
client in 'cargo new rust-client' or 'cargo new rust-client'

finally the dir is
```
[workspace]

members = [
    "lib-geyser-grpc",
    "rust-client",
]

/Cargo.toml
/lib-geyser-grpc/Cargo.toml
/lib-geyser-grpc/src/
/rust-client/Cargo.toml
/rust-client/src/client.rs
```


fanatid:
Thank you that re-organizing the project, it's good to have the client as "separate" project because it's clear how gRPC can be used without looking back to the plugin code. I think it would be better to have directory `examples` with js/ts/golang/rust but we can make it in the future if some other directory changes will be introduced.

PR #72, opened 2023-02-23
fix: x-token in client is optional

shuimuliang:
to make local test convenient:
if x-token is empty, will show a warning message.
if x-token is not empty, will check XTOKEN_LENGTH == 28.

### DIFF
--- a/rust/client/client.rs
+++ b/rust/client/client.rs
@@ -16,6 +16,8 @@ use {
         transport::{channel::ClientTlsConfig, Channel, Endpoint},
         Request, Response, Status,
     },
+    log::{info, warn, error},
+    env_logger,
 };
 
 #[derive(Debug, Parser)]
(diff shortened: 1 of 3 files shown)


rust/client/client.rs
```diff
@@ -209,26 +214,29 @@ impl<F: FnMut(Request<()>) -> InterceptedRequestResult> RetryClient<F> {
...
         }
-        println!("stream closed");
+        info!("stream closed");
         Ok(())
     }
 }
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
+    ::std::env::set_var("RUST_LOG", "info");
```
fanatid, line 237: is it good to redefine `RUST_LOG`? IIRC if env is not set then `info` used by default?

fanatid:
LGTM, thanks!

rust/client/client.rs
georger-fm: https://docs.rs/env_logger/latest/env_logger/
Log levels are controlled on a per-module basis, and by default all logging is disabled except for the error level.

so enabled "info".

PR #85, opened 2023-03-09
Don't close streams

linuskendall:
Removes the work arounds for the bidi streams not working as expected due to our loadbalancer bug.


golang/cmd/grpc-client/main.go
```diff
@@ -232,9 +232,18 @@ func grpc_subscribe(conn *grpc.ClientConn) {
 	if err != nil {
 		log.Fatalf("%v", err)
 	}
-	stream.CloseSend()
 
+  i := 0
```
fanatid: Do we really need this counter?

linuskendall: No, before merging we should come up with some nice way to demonstrate re-subs in the clients?

fanatid: maybe some way as in rust example? left counter but comment line where we send new subscribe config?

linuskendall: I've added a new resub param. Good to keep this option for testing and example illustration.

rust/client/client.rs
```diff
@@ -295,6 +304,11 @@ async fn main() -> Result<(), Error> {
         blocks_meta.insert("client".to_owned(), SubscribeRequestFilterBlocksMeta {});
     }
 
+    let mut resub: u16 = 0;
```
fanatid, line 307:
```
let resub = args.resub.unwrap_or(0)
```
:slightly_smiling_face:

PR #79, opened 2023-02-23
Update solana =1.15.2

fanatid:
Closes #61

@linuskendall any ideas on how I can update golang proto files? Currently, I get an error:
```
$ make protoc 
protoc --go_out=./proto --go_opt=paths=source_relative --go-grpc_out=./proto --go-grpc_opt=paths=source_relative --proto_path ../proto/ ../proto/*.proto
protoc-gen-go-grpc: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--go-grpc_out: protoc-gen-go-grpc: Plugin failed with status code 1.
make: *** [Makefile:2: protoc] Error 1
```

PR #86, opened 2023-03-09
Remove old golang proto

linuskendall:
Remove duplicate golang proto.


fanatid:
Thanks, #79 was not accurate :disappointed:

PR #99, opened 2023-03-21
update to geyser interface with less locks

fanatid:
Looks like a patch with fewer locks in the geyser introduce some problems to `validator`. `owner` size in the account info on `update_account` is not 32 because of this I get a segmentation fault. Maybe it will be good to wait a while. We can re-use this code when `validator` with a changed geyser interface will be released.


fanatid:
- outdated
- we already started testing 1.16 (`&self` used instead of `&mut self`)

PR #117, opened 2023-05-08
114 the grpc framework supports non streaming methods getlatestblockhash

shuimuliang:


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -25,6 +26,7 @@ use {
     std::{
         collections::HashMap,
         sync::atomic::{AtomicUsize, Ordering},
+        sync::{Arc, RwLock},
```
fanatid: I think it's better to use RwLock from tokio, everything what we have is async

shuimuliang: notify_block_metadata is not async fn, if call await in it
```rust
    if let Ok(mut latest_block_meta_write_guard) = inner.latest_block_meta.write().await {
        *latest_block_meta_write_guard = Some(block_meta.clone());
    }
```
error[E0728]: `await` is only allowed inside `async` functions and blocks

shuimuliang:
for your review @fanatid

yellowstone-grpc-geyser/src/grpc.rs
fanatid: oh, I missed that we set latest value in the sync function, so use std RwLock was good idea. Sorry.

fanatid: https://github.com/rpcpool/yellowstone-grpc/pull/118

Issue #112, opened 2023-04-20
Add non-streaming methods


linuskendall:
I would like to add non streaming methods that could help users who are writing trading code:

- [x] getLatestBlockhash
- [x] getBlockHeight
- [x] getSlot (we already have support ?)

Harder ones that we could potentially do if we add much more logic to the code
- [ ] getRecentPrioritizationFee (need to store transcation stats, probably out-of-scope)
- [x] isBlockhashValid (need to store block hashes)

Ones that I don't know if possible (what info would we require to do this? maybe some modification to geyser interface to send some of this info on start up):
- [x] getVersion 
- [ ] getIdentity
- [ ] getEpochInfo

Others..?

linuskendall:
See also #47, which can be a first version (not sure if it can include the solana validator version?).

shuimuliang:
https://docs.solana.com/api/http#getlatestblockhash
https://docs.solana.com/api/http#getblockheight
https://docs.solana.com/api/http#getslot

shuimuliang:
linuskendall:
for getLatestBlockhash
if we can we should use the existing interface
I think via block notification + slot notification

shuimuliang:
share read/write design for MessageBlockMeta

```rust
// plugin.rs
#[derive(Debug)]
pub struct PluginInner {
    // ...
    latest_block_meta: Arc<RwLock<Option<MessageBlockMeta>>>,
}

// grpc.rs
#[derive(Clone)]
pub struct GrpcService {
    // ...
    latest_block_meta: Arc<RwLock<Option<MessageBlockMeta>>>,
}

// plugin.rs
    fn notify_block_metadata(
        &mut self,
        blockinfo: ReplicaBlockInfoVersions<'_>,
    ) -> PluginResult<()> {
                // ...
               if let Ok(mut latest_block_meta_write_guard) = inner.latest_block_meta.write() {
                *latest_block_meta_write_guard = Some(block_meta.clone());
            }
}

// grpc.rs
    async fn get_latest_blockhash(&self, _request: Request<GetLatestBlockhashRequest>) -> Result<Response<GetLatestBlockhashResponse>, Status> {
        if let Ok(message_block_meta) = self.latest_block_meta.read() {
            let v= message_block_meta.as_ref();
            if let Some(v) = v {
                let response = GetLatestBlockhashResponse {
                    slot: v.slot,
                    blockhash: v.blockhash.clone(),
                    block_height: v.block_height.unwrap(),
[...]

PR #122, opened 2023-05-22
rust client: add unary methods

shuimuliang:
add ping,get_latest_blockhash,get_block_height,get_slot for yellowstone-grpc-client

### DIFF
--- a/examples/rust/src/bin/client.rs
+++ b/examples/rust/src/bin/client.rs
@@ -9,10 +9,17 @@ use {
     yellowstone_grpc_proto::prelude::{
         subscribe_request_filter_accounts_filter::Filter as AccountsFilterDataOneof,
         subscribe_request_filter_accounts_filter_memcmp::Data as AccountsFilterMemcmpOneof,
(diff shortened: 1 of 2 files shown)


yellowstone-grpc-client/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+use yellowstone_grpc_proto::geyser::{GetSlotResponse, PongResponse};
+use yellowstone_grpc_proto::prelude::{GetBlockHeightResponse, GetLatestBlockhashResponse};
```
fanatid, line 2: we have import from this crate down in this file (line 19)

examples/rust/src/bin/client.rs
```diff
@@ -288,3 +295,17 @@ async fn main() -> anyhow::Result<()> {
     .await
     .map_err(Into::into)
 }
+
+// unary method
```
fanatid, line 299: I think it would be better support unary methods in arguments, not as commented code

fanatid:
Related #112

PR #128, opened 2023-05-23
grpc, proto: add commitment level

### LINKED ISSUE #20: Transaction commitment
### Problem

I am using the gRPC geyser service as data source to index transactions, and for that it is important I trust that the transactions I receive actually make it into the block. However I notice that my DB has transactions that cannot be found on-chain.

My understanding from our discussion is that the intent is to let users cross-reference the transaction sub updates with the slot sub updates (giving the corresponding slot commitment) client-side, to know the txs' commitments. However afaik this makes the assumption that no block is uncled == any transaction received (i.e. `processed`) "makes it".

At the moment I need to perform an unsightly verification step to avoid storing wrong data, which is not ideal.

### Proposal

I haven't dug into the server codebase, but I'm hoping there is a way for you to move the commitment matching to the source, therefore allowing transaction subscriptions with a specific commitment.

And let me know if I simply oversaw something, of course 😄

fanatid:
Closes #20

### DIFF
--- a/examples/rust/src/bin/client.rs
+++ b/examples/rust/src/bin/client.rs
@@ -150,7 +150,7 @@ struct ActionSubscribe {
 impl Action {
     fn get_subscribe_request(
         &self,
-        commitment: Option<i32>,
+        commitment: Option<CommitmentLevel>,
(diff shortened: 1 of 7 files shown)


fanatid:
Approved by @shuimuliang in DM.

Issue #123, opened 2023-05-22
implement the isBlockhashValid method


shuimuliang:

shuimuliang:
A valid blockhash basically means "Can I use this blockhash to send a transaction?”

it's means that we do not need any database, we can store hashes in memory, only need to know the size of stored queue.

default value is 300

[https://docs.rs/solana-sdk/latest/solana_sdk/clock/constant.MAX_RECENT_BLOCKHASHES.html](https://docs.rs/solana-sdk/latest/solana_sdk/clock/constant.MAX_RECENT_BLOCKHASHES.html)

shuimuliang:
Linus:

I think it’s ok for server start up that we return unknown

so if someone requests isBlockhashValid when we have less < 300 blockhashes in the queue

then we can just answer “unknown”

it will only take 5 minutes from start up until we have all the known blockhashes in store

PR #129, opened 2023-05-24
feat: implement the isBlockhashValid method #123

shuimuliang:
if someone requests isBlockhashValid when we have less < 300 blockhashes in the queue then we can just answer “we have less < 300 blockhashes in the queue”


fanatid:
Replaced by https://github.com/rpcpool/yellowstone-grpc/pull/132

PR #142, opened 2023-06-14
Add TypeScript client package

nbaksalyar:
- TypeScript client is now packaged as a library.
- Add more supported commands in the TypeScript example.
- Conform the TS client API with the Rust API.


yellowstone-grpc-client-nodejs/src/grpc/.gitignore
fanatid, line 1: do we really need this file? or used to keep a directory, like `.gitkeep`?

yellowstone-grpc-client-nodejs/src/index.ts
```diff
@@ -0,0 +1,193 @@
...
+        (err) => {
+          if (err === null || err === undefined) {
+            resolve();
+          } else {
+            reject(err);
+          }
+        }
+      );
+    }).catch((reason) => {
+      console.error(reason);
```
fanatid: is it good to print error in the client library?

examples/typescript/README.md
```diff
@@ -1,26 +1,128 @@
...
+npm start -- --endpoint "https://api.rpcpool.com" \
+  --x-token "<token>" \
+  subscribe \
+  --accounts --accounts-account "<Pubkey>"
+```
+
+### subscribe to slot updates
+
+```shell
+npm start -- --endpoint https://api.rpcpool.com \
```
fanatid, line 26: in the previous example, we have extra quotes around the endpoint, but here no -- is it intentional to show that different variants are accepted? just curious

fanatid (approved):
utACK

yellowstone-grpc-client-nodejs/src/index.ts
nbaksalyar: Good catch, thanks :+1:

yellowstone-grpc-client-nodejs/src/grpc/.gitignore
nbaksalyar: It's not strictly required to be there. Moved it to the parent directory. :+1:

examples/typescript/README.md
nbaksalyar: Both variants are accepted, but you're right - I updated it for consistency.

yellowstone-grpc-client-nodejs/.npmignore
```diff
@@ -0,0 +1 @@
+node_modules
```
fanatid, line 1: it can be worth using `files` in `package.json ` instead of `.npmignore` file, whitelist is always better for packages?
https://docs.npmjs.com/cli/v9/configuring-npm/package-json#files

fanatid (approved):
LGTM, thanks!

yellowstone-grpc-client-nodejs/.npmignore
nbaksalyar: Sounds good, will make the change in one of the next pull requests. Thanks!

PR #148, opened 2023-06-20
2204 workflow

Juanito87:


Lusitaniae:
Some notes so far:
- Using matrix to run the job in multiple environments (earlier condition was ubuntu20 AND ubuntu22 filter, which didn't find any runner for that combination)
- Merged all into the same job because it's wasn't passing files around jobs; and it requires env variables set in the first job on the 2nd and 3rd too

Juanito87:
We are going to use the changes implemented on #153

PR #153, opened 2023-06-28
ci: add ubuntu-22

fanatid:


Juanito87 (commented):
In the release action, the last 2 lines (85-86) should have a '*' at the end of the line, so the release will update both versions of the binaries (client and config-check)

fanatid:
> In the release action, the last 2 lines (85-86) should have a '*' at the end of the line, so the release will update both versions of the binaries (client and config-check)

Thanks!

PR #159, opened 2023-07-14
Updating workflow.

Juanito87:


.github/workflows/release.yml
```diff
@@ -63,10 +74,9 @@ jobs:
         run: ./ci/create-tarball.sh
 
       - name: rename binaries for ubuntu22 release
-        if: matrix.os == 'ubuntu-22.04'
```
fanatid, line 66: why this condition is removed?

Juanito87: Good catch, that shouldn't have happened.

fanatid: ok, but if we rename `client`/`config-check` only for ubuntu-22 then on the upload pattern `client-*`/`config-check-*` will not match anything?

Juanito87: Yeah, you are right. SO we have two options:

* We also rename the 20 files, so we can get a match
* we have two separated release actions, one for each os version (into the same release) to upload specific files

I think that the later option will be the cleanest.

fanatid: What if we add an additional step where we set some env variable to `20` or `22` (depends on `matrix.os`) and will use this env to rename files?

Juanito87: That would be the first option, we set the os version as part of the name, duplicating this step.so it would be 20 or 22 dependding on which os is build, but this would also will create an issue on the rpcpool repo.
On the other hand,  I think that we could keep things as is, and instead of working a way to use the limit glob we have for uploads, we can simply rename the folder that causes issues with the glob. LIke:

`mv config-check.d aconfig0check.d`
This way seems to be the simplest approach, and shouldn't have any effect on the build.

fanatid: Yes, indeed, we can rename/remove `.d` files :) sounds good to me :+1:

.github/workflows/release.yml
```diff
@@ -65,11 +76,16 @@ jobs:
...
         if: matrix.os == 'ubuntu-22.04'
         run: |
-          mv target/release/client target/release/client22
-          mv target/release/config-check target/release/config-check22
+          mv target/release/client target/release/client-22
+          mv target/release/config-check target/release/config-check-22
           mv ${{ env.GEYSER_PLUGIN_NAME }}-release-x86_64-unknown-linux-gnu.tar.bz2 ${{ env.GEYSER_PLUGIN_NAME }}-release22-x86_64-unknown-linux-gnu.tar.bz2
           mv ${{ env.GEYSER_PLUGIN_NAME }}-release-x86_64-unknown-linux-gnu.yml ${{ env.GEYSER_PLUGIN_NAME }}-release22-x86_64-unknown-linux-gnu.yml
 
+      - name: rename binaries for ubuntu22 release
```
fanatid: step name looks a little confused :upside_down_face: 
instead of renaming, we can remove `.d` in one command
LGTM

PR #167, opened 2023-07-24
add full block notification with no account defined.

musitdev:
I change the filter check_any verification to allow block filter with no account defined but Tx or Account activated.
It allows to be notified for each full  block and its Tx and/or accounts.
We need it for the RPCv2.

I update the default config.json to add the mandatory block fields if block notification is activated.

### DIFF
--- a/yellowstone-grpc-geyser/config.json
+++ b/yellowstone-grpc-geyser/config.json
@@ -32,7 +32,9 @@
                 "max": 1,
                 "any": false,
                 "account_include_max": 10,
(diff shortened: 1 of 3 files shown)


fanatid:
Thank you for PR! I submit small changes to you PR here: https://github.com/blockworks-foundation/yellowstone-grpc/pull/1

musitdev:
I've tested with your modification, and it's working for me. You can merge when you prefer.

fanatid:
Thank you, merged in #168

PR #207, opened 2023-10-16
bug(plugin): signal end of startup for genesis cluster if parent slot 0 seen

vovkman:
if plugin run on genesis of network, `notify_end_of_startup` will not be triggered. Slot 0 will not come through `update_slot_status` so the earliest signal is parent slot 0


yellowstone-grpc-geyser/src/plugin.rs
```diff
@@ -174,6 +174,13 @@ impl GeyserPlugin for Plugin {
         status: SlotStatus,
     ) -> PluginResult<()> {
         let inner = self.inner.as_ref().expect("initialized");
+        // if plugin run on genesis of network, notify_end_of_startup will not be triggered, and slot 0 will not
+        // come through this function.
+        if parent.eq(&Some(0)) {
```
fanatid: does `.eq` have any benefits instead of `==`?
```suggestion
        if parent == Some(0) {
```

fanatid (approved):
Awesome! This would allow us to run the plugin without pre-generated `test-ledger` with a full snapshot.
Can you please add a line to `CHANGELOG.md`?

yellowstone-grpc-geyser/src/plugin.rs
vovkman: Sometimes compiler complains, but this one doesn't seem to matter so I'll update

PR #236, opened 2023-11-15
fixes openssl link problem (on mac os)

grooviegermanikus:
### Problem
building the _rdkafka_ crate fails on mac per default configuration

>   _mkltmpuU0ERk.c:2:10: fatal error: 'openssl/ssl.h' file not found
  #include <openssl/ssl.h>
           ^~~~~~~~~~~~~~~
  1 error generated.

see https://github.com/fede1024/rust-rdkafka/issues/491

### Proposed solution:
ssl -> ssl-vendored


fanatid:
Thank you!

PR #242, opened 2023-11-19
Use release build

RadiantAeon:


fanatid:
This is just an example, I do not think it matters a lot what build profile is actually set here

PR #252, opened 2023-12-04
fix(client): Include request in initial subscribe to gRPC endpoint to fix LB connection delay

vovkman:
Some load balancers (eg. Cloudflare) expect a message to come through before passing a gRPC connection through to origin. Because the current subscribe method connects, and then sends a message, it causes a delay when connecting. This pr adds `subscribe_with_request` which pushes a request into the channel before sending the subscribe request. This resolves the LB issue


fanatid:
Thanks!

PR #254, opened 2023-12-05
chore(client): Update readme and rust example

vovkman:
In relation to #252 load balancers sometimes require that client sends a ping. Updating the readme to call this out and including an example in the rust examples

### DIFF
--- a/examples/rust/src/bin/client.rs
+++ b/examples/rust/src/bin/client.rs
@@ -524,11 +524,7 @@ async fn geyser_subscribe(
     request: SubscribeRequest,
     resub: usize,
 ) -> anyhow::Result<()> {
-    let (mut subscribe_tx, mut stream) = client.subscribe().await?;
-    subscribe_tx
-        .send(request)
-        .await
-        .map_err(GeyserGrpcClientError::SubscribeSendError)?;
+    let (mut subscribe_tx, mut stream) = client.subscribe_with_request(Some(request)).await?;
 
     info!("stream opened");
     let mut counter = 0;
--- a/examples/rust/src/bin/client.rs
+++ b/examples/rust/src/bin/client.rs
@@ -524,7 +524,8 @@ async fn geyser_subscribe(
     request: SubscribeRequest,
     resub: usize,
 ) -> anyhow::Result<()> {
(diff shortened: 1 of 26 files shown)


examples/rust/src/bin/client.rs
```diff
@@ -549,6 +550,23 @@ async fn geyser_subscribe(
...
+                            .send(SubscribeRequest {
+                                ping: Some(SubscribeRequestPing { id: 1 }),
+                                slots: request.slots.clone(),
+                                accounts: request.accounts.clone(),
+                                transactions: request.transactions.clone(),
+                                entry: request.entry.clone(),
+                                blocks: request.blocks.clone(),
+                                blocks_meta: request.blocks_meta.clone(),
+                                commitment: request.commitment.clone(),
+                                accounts_data_slice: request.accounts_data_slice.clone(),
```
fanatid:
```suggestion
                                ping: Some(SubscribeRequestPing { id: 1 }),
                                ..Default::default()
```
if `ping` field is defined all other fields are ignored, so we can set them to default value

vovkman: Oh right I was testing on a branch that didn't have the latest ping changes

vovkman: Updated, sorry for delay

fanatid: np, thanks for the update!

examples/rust/src/bin/client.rs
```diff
@@ -524,7 +524,8 @@ async fn geyser_subscribe(
     request: SubscribeRequest,
     resub: usize,
 ) -> anyhow::Result<()> {
-    let (mut subscribe_tx, mut stream) = client.subscribe_with_request(Some(request)).await?;
+    let (mut subscribe_tx, mut stream) =
+        client.subscribe_with_request(Some(request.clone())).await?;
```
fanatid:
```suggestion
        client.subscribe_with_request(Some(request)).await?;
```

`.clone()` is not required now

vovkman: updated

fanatid:
Thank you!

PR #287, opened 2024-02-19
feat: Accounts from path

Arrowana:
Allow testing with very large filters, like jupiter-swap-api


fanatid:
small async update https://github.com/jup-ag/yellowstone-grpc/pull/1

fanatid:
merged in #290

PR #295, opened 2024-03-07
up max_decoding_message_size to 8_388_608

stakeconomy:


fanatid:
4mb as value for example is good, its default value in tonic

PR #309, opened 2024-03-30
client: add `GeyserGrpcBuilder`

fanatid:


grooviegermanikus (commented):
lgtm.
not sure is you use semantic versioning. the change could be non-breaking, right?

fanatid:
strictly speaking yes, PR introduces breaking change and I made only a minor version bump (does solana do the same with geyser interface :zany_face: ?)
probably yes, would be good to do major version bump :thinking:

vovkman:
Just wondering if there is a specific reason to not make it backwards compatible?

grooviegermanikus (approved):

PR #310, opened 2024-03-30
geyser: add `transactions_status` filter

fanatid:
@vovkman @NicolasPennie @godmodegalactus @grooviegermanikus what do you think about a new message for Atlas / Lite RPC? We do not need to receive all transaction details to know that transaction confirmed or not.

```proto
message SubscribeUpdateTransactionStatus {
  uint64 slot = 1;
  bytes signature = 2;
  bool is_vote = 3;
  uint64 index = 4;
  solana.storage.ConfirmedBlock.TransactionError err = 5;
}
```


vovkman:
We sub to all blocks so we wouldn't use this in atlas, but I could see this being useful for lots of other cases!

examples/rust/src/bin/client.rs
```diff
@@ -162,6 +163,34 @@ struct ActionSubscribe {
...
+    #[clap(long)]
+    transactions_status_vote: Option<bool>,
+
+    /// Filter failed transactions for transactions_status
+    #[clap(long)]
+    transactions_status_failed: Option<bool>,
+
+    /// Filter by transaction signature for transactions_status
+    #[clap(long)]
+    transactions_status_signature: Option<String>,
```
godmodegalactus, line 180: Should be more like a vector of string, so that filtering can be done on multiple signatures in one subscription..

godmodegalactus (changes requested):

examples/rust/src/bin/client.rs
fanatid: good point but unfortunately in transactions filter we have signature as `optional string` not as `repeated string` :confused:

PR #318, opened 2024-04-04
Add grpc2scylla and scylladb module

lvboudre:


yellowstone-grpc-tools/Cargo.toml
```diff
@@ -60,3 +65,4 @@ vergen = { workspace = true, features = ["build", "rustc"] }
 default = ["google-pubsub", "kafka"]
 google-pubsub = ["google-cloud-googleapis", "google-cloud-pubsub"]
 kafka = ["const-hex", "rdkafka", "sha2"]
+#scylladb = ["scylla"]
```
fanatid: is there some reason do not have `scylladb` as a feature?

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,154 @@
...
+           sink::ScyllaSink,
+           config::Config
+        },
+        prom::{run_server as prometheus_run_server},
+        setup_tracing,
+    },
+};
+
+#[derive(Debug, Clone, Parser)]
+#[clap(author, version, about = "Yellowstone gRPC Kafka Tool")]
```
fanatid:
```suggestion
#[clap(author, version, about = "Yellowstone gRPC Scylladb Tool")]
```

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,154 @@
...
+
+
+
+impl ArgsAction {
+    async fn run(self, config: Config) -> anyhow::Result<()> {
+        let shutdown = create_shutdown()?;
+        match self {
+            ArgsAction::Grpc2Scylla => {
+                let config2 = config.grpc2scylladb.ok_or_else(|| {
+                    anyhow::anyhow!("`grpc2kafka` section in config should be defined")
```
fanatid:
```suggestion
                    anyhow::anyhow!("`grpc2scylladb` section in config should be defined")
```

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,154 @@
...
+    ) -> anyhow::Result<()> {
+        let sink_config = ScyllaSinkConfig {
+            batch_size_limit: config.batch_size_limit,
+            linger: config.linger,
+            keyspace: config.keyspace,
+            max_inflight_batch_delivery: config.max_inflight_batch_delivery,
+        };
+
+        // Create gRPC client & subscribe
+        let mut client = GeyserGrpcClient::connect_with_timeout(
```
fanatid: The subscription code changed, please check kafka tool

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,154 @@
...
+                Some(message) => {
+                    let message = match message.update_oneof {
+                        Some(value) => value,
+                        None => unreachable!("Expect valid message"),
+                    };
+                    match message {
+                        UpdateOneof::Account(msg) => {
+                            sink.log_account_update(msg.into()).await;
+                        },
+                        _ => continue,
```
fanatid, line 143: if we do not need other messages then `if let UpdateOneof::Account(msg) = ...` would be better

yellowstone-grpc-tools/src/scylladb/config.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[serde(default = "default_hostname")]
+    pub hostname: String
+}
+
+#[derive(Debug, Deserialize)]
+pub struct ConfigGrpc2ScyllaDB {
+    pub endpoint: String,
+    pub x_token: Option<String>,
+    pub request: ConfigGrpcRequest,
+    #[serde(default = "default_batch_size_limit")]
```
fanatid: can we move default functions into `ConfigGrpc2ScyllaDB`? like kafka/pubsub

yellowstone-grpc-tools/src/scylladb/types.rs
```diff
@@ -0,0 +1,64 @@
...
+            self.slot, 
+            self.pubkey, 
+            Default::default(), 
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default()
+        )
```
fanatid, line 473: `Self { slot: self.slot, pubkey: self.pubkey, ..Default::default() }.as_row()` should work

yellowstone-grpc-tools/src/scylladb/types.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl From<SubscribeUpdateAccount> for AccountUpdate {
+    fn from(value: SubscribeUpdateAccount) -> Self {
+        let slot = value.slot;
+        let acc: yellowstone_grpc_proto::prelude::SubscribeUpdateAccountInfo = value.account.unwrap();
+        AccountUpdate {
+            slot: slot as i64,
+            pubkey: acc.pubkey.try_into().unwrap(),
+            lamports: acc.lamports as i64,
+            owner: acc.owner.try_into().unwrap(),
```
fanatid: `.unwrap()` is not the best thing, can we change from `From` to `TryFrom`?

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,367 @@
...
+        
+        (ScyllaSink { insert_account_update_ps, batcher: batcher, batcher_handle, error_sender: sender }, receiver)
+    }
+
+    pub async fn log_account_update(&mut self, update: AccountUpdate) {
+        let br = BatchRequest {
+            ps: self.insert_account_update_ps.clone(),
+            item: BatchItem::Account(update)
+        };
+        self.batcher.send(br).unwrap();
```
fanatid: please return an error instead `.unwrap()`

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,154 @@
...
+        config: ConfigGrpc2ScyllaDB,
+        scylladb_conn_config: ScyllaDbConnectionInfo,
+        mut shutdown: BoxFuture<'static, ()>,
+    ) -> anyhow::Result<()> {
+        let sink_config = ScyllaSinkConfig {
+            batch_size_limit: config.batch_size_limit,
+            linger: config.linger,
+            keyspace: config.keyspace,
+            max_inflight_batch_delivery: config.max_inflight_batch_delivery,
+        };
```
fanatid: we can add a function to `ConfigGrpc2ScyllaDB` that create `ScyllaSinkConfig`

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,367 @@
...
+        tokio::select! {
+            _ = time::sleep_until(deadline), if !self.throttle()  => {
+                let old_value = self.consume_and_refresh_next_in_schedule(now);
+                if let Some((batch, ser_values)) = old_value {
+                    let bs = self.batch_sender.clone();
+                    let fut = bs.send_batch(batch, ser_values);
+                    self.inflight_deliveries.spawn(fut);
+                }
+            },
+            opt_msg = self.inner.recv(), if !self.throttle() => {
```
fanatid: I do not think that this is correct, `if !self.throttle()` is applied when `self.inner.recv()` resolved, so the received message would be dropped
if you want throttle then you need to create future, like
```
let inner_recv = if self.throttle() { pending().boxed() } else {self.inner.recv().boxed()};
```

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,367 @@
...
+    }
+    
+    fn compute_token(&self, table: &str, partition_key: &SerializedValues) -> Token {
+        let current_keysapce = self.0.get_keyspace().unwrap();
+        self.0.get_cluster_data().compute_token(&current_keysapce, table, partition_key).unwrap()
+    }
+}
+
+#[async_trait]
+pub trait BatchSender: Send + Sync + Clone {
```
fanatid: do we need a trait here? would be much easier to implement as method, also static dispatch should be more efficient

fanatid (commented):
I know this is a draft but we need:
- [x] metrics
- [x] remove unwrap and add errors
- [x] some logs would be good, at least on debug level
- [x] cargo fmt
- [x] cargo clippy

yellowstone-grpc-tools/Cargo.toml
lvboudre: I forgot to uncommment it. I was having issue with rust-analyzer in Vscode :(

yellowstone-grpc-tools/src/scylladb/sink.rs
lvboudre: Good catch! I've read the code of tokio::select!, indeed the futures gets created but not pulled if the precondition is false, thus there is a possibility that we can lose a message.

yellowstone-grpc-tools/src/scylladb/sink.rs
lvboudre: It was for testability purposes when unit testing the batcher where I don't want to actually send it to Scylladb.
Otherwise my Batcher would be coupled with a scyllaDB `Session` which is hard to test with.

For the static dispatch, since my TokenAwareBatcher has a generic parameter for BatchSender, there will be a compiled version of the function with static dispatch for each BatchSender that the compiler detects.

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
lvboudre: Got it! However I will need to impleement transaction as well ;)

.github/workflows/test.yml
```diff
@@ -36,20 +36,18 @@ jobs:
           echo "GEYSER_PLUGIN_NAME=$plugin_name" | tee -a $GITHUB_ENV
           echo "GEYSER_PLUGIN_LIB=lib${plugin_lib_name}" | tee -a $GITHUB_ENV
 
-      - uses: actions-rs/toolchain@v1
-        with:
-          toolchain: nightly
-          override: true
-          profile: minimal
-          components: rustfmt
-
```
fanatid, line 43: nightly required for `cargo +nightly fmt`

Juanito87: Yeah, just bumped into that and reverted.
When I have some time, i would like to see if there is a better way manage workflows and build pipelines for this repos.

.github/workflows/test.yml
```diff
@@ -34,18 +34,14 @@ jobs:
...
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: nightly
-          override: true
-          profile: minimal
           components: rustfmt
 
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
fanatid: can we move unrelated ci changed to separate pr?
btw, I think better would be to use `@v1` (at least I use v1 in all other repos)

Cargo.toml
```diff
@@ -68,6 +68,9 @@ tracing-subscriber = "0.3.17"
 vergen = "8.2.1"
 yellowstone-grpc-client = { path = "yellowstone-grpc-client", version = "=1.15.0+solana.1.18.11" }
 yellowstone-grpc-proto = { path = "yellowstone-grpc-proto", version = "=1.14.0+solana.1.18.11" }
+scylla = "0.12.0"
+serde_with = "3.7.0"
+deepsize = "0.2.0"
```
fanatid: can you please place dependencies in alphabetic order? easier to review the list in the future

yellowstone-grpc-tools/config-scylladb.json
```diff
@@ -0,0 +1,23 @@
+{
+    "prometheus": "127.0.0.1:8873",
+    "scylladb": {
+        "hostname": "localhost:9042",
+        "username": "cassandra",
+        "password": "cassandra"
+    },
+    "grpc2scylladb": {
+        "endpoint": "https://index.rpcpool.com:10000",
```
fanatid: in config example probably would be better to point on localhost:10000

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,177 @@
...
+            config::{Config, ConfigGrpc2ScyllaDB, ScyllaDbConnectionInfo},
+            sink::ScyllaSink,
+            types::{Reward, Transaction},
+        },
+        setup_tracing,
+    },
+};
+
+// 512MB
+const MAX_DECODING_MESSAGE_SIZE_BYTES: usize = 512000000;
```
fanatid:
```suggestion
const MAX_DECODING_MESSAGE_SIZE_BYTES: usize = 512_000_000;
```

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,177 @@
...
+        let shutdown = create_shutdown()?;
+        match self {
+            ArgsAction::Grpc2Scylla => {
+                let config2 = config.grpc2scylladb.ok_or_else(|| {
+                    anyhow::anyhow!("`grpc2scylladb` section in config should be defined")
+                })?;
+                Self::grpc2scylladb(config2, config.scylladb, shutdown).await
+            }
+            ArgsAction::Scylla2Grpc => {
+                Ok(())
```
fanatid: should we print some error that the command is not implemented yet? or even https://doc.rust-lang.org/std/macro.unimplemented.html

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,177 @@
...
+
+                    match message {
+                        UpdateOneof::Account(msg) => {
+                            // let pubkey_opt: &Option<&[u8]> =
+                            //     &msg.account.as_ref().map(|acc| acc.pubkey.as_ref());
+                            // trace!(
+                            //     "Received an account update slot={:?}, pubkey={:?}",
+                            //     msg.slot, pubkey_opt
+                            // );
+                            let acc_update = msg.clone().try_into();
```
fanatid, line 123: why do we need clone here?

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -0,0 +1,177 @@
...
+                            if acc_update.is_err() {
+                                // Drop the message if invalid
+                                warn!(
+                                    "failed to parse account update: {:?}",
+                                    acc_update.err().unwrap()
+                                );
+                                continue;
+                            }
+                            // If the sink is close, let it crash...
+                            sink.log_account_update(acc_update.unwrap()).await.unwrap();
```
fanatid, line 133:
```suggestion
                            match msg.try_into() {
                                Err(error) => {
                                    // Drop the message if invalid
                                    warn!("failed to parse account update: {error:?}");
                                }
                                Ok(acc_update) => {
                                    // If the sink is close, let it crash...
                                    sink.log_account_update(acc_update.unwrap()).await?;
                                }
                            }
```

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,953 @@
...
+        prom::{
+            scylladb_batch_request_lag_inc, scylladb_batch_request_lag_sub,
+            scylladb_batch_sent_inc, scylladb_batch_size_observe, scylladb_batchitem_sent_inc_by,
+        },
+        types::{
+            ProducerId, ShardId, ShardOffset, ShardPeriod, ShardStatistics, ShardedAccountUpdate,
+            Transaction, SHARD_OFFSET_MODULO,
+        },
+    },
+    crate::scylladb::{
```
fanatid: `super::` and `crate::scalladb::` can be merged, right?

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,953 @@
...
+    );
+    Ok(shard)
+}
+
+impl ScyllaSink {
+    pub async fn new(
+        config: ScyllaSinkConfig,
+        hostname: impl AsRef<str>,
+        username: impl Into<String>,
+        password: impl Into<String>,
```
fanatid, line 801: why do not pass config instead of hostname / username / password?

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,953 @@
...
+        let producer_id = [config.producer_id];
+
+        let session: Session = SessionBuilder::new()
+            .known_node(hostname)
+            .user(username, password)
+            .compression(Some(Compression::Lz4))
+            .use_keyspace(config.keyspace.clone(), false)
+            .build()
+            .await
+            .unwrap();
```
fanatid, line 812: please, remove unwrap, anyhow::Context also would be a good addition

yellowstone-grpc-tools/src/scylladb/types.rs
```diff
@@ -0,0 +1,705 @@
...
+    pub stack_height: Option<i64>,
+}
+
+impl From<confirmed_block::InnerInstruction> for InnerInstr {
+    fn from(value: confirmed_block::InnerInstruction) -> Self {
+        InnerInstr {
+            program_id_index: value.program_id_index.into(),
+            accounts: value.accounts,
+            data: value.data,
+            stack_height: value.stack_height.map(|x| x.into()),
```
fanatid, line 238:
```suggestion
            stack_height: value.stack_height.map(Into::into),
```

yellowstone-grpc-tools/src/scylladb/types.rs
```diff
@@ -0,0 +1,705 @@
...
+    I::Item: TryInto<U> + fmt::Debug,
+    <I::Item as TryInto<U>>::Error: fmt::Debug,
+{
+    let mut res = Vec::new();
+    for x in it.into_iter() {
+        let y = x.try_into();
+        if y.is_err() {
+            return Err(y.unwrap_err());
+        }
+        res.push(y.unwrap());
```
fanatid:
```suggestion
        res.push(x.try_into()?);
```
I do not understand why clippy does not catch it, if we can easily remove unwrap we **should** remove unwrap

even more, everything can be mapped like this:
```rust
    it.into_iter()
        .map(|item| item.try_into())
        .collect::<Result<Vec<_>, _>>()
```

yellowstone-grpc-tools/src/scylladb/agent.rs
```diff
@@ -0,0 +1,331 @@
...
+                            },
+                        }
+                    }
+                };
+
+                if result.is_err() {
+                    let emsg = format!("message loop: {:?}, {:?}", name, result.err().unwrap());
+                    error!(emsg);
+                    return Err(anyhow!(emsg));
+                }
```
fanatid:
```suggestion
                if let Err(error) = result {
                    let emsg = format!("message loop: {name:?}, {error:?}");
                    error!(emsg);
                    anyhow::bail!(emsg);
                }
```
should we log if return error? :thinking:

yellowstone-grpc-tools/src/scylladb/agent.rs
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    pub fn until_one_agent_dies(&mut self) -> impl Future<Output = anyhow::Result<Nothing>> + '_ {
+        self.handlers
+            .join_next()
+            .map(|inner| inner.unwrap_or(Ok(Ok(()))))
+            .map(|result| match result {
+                Ok(result2) => result2,
+                Err(e) => Err(anyhow::Error::new(e)),
+            })
```
fanatid:
```suggestion
            .map(|result| result.map_err(Into::into))
```

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -0,0 +1,953 @@
...
+    }
+
+    async fn tick_with_callback_sender(
+        &mut self,
+        now: Instant,
+        msg: Self::Input,
+        callback_senders: Vec<CallbackSender>,
+    ) -> anyhow::Result<Nothing> {
+        self.tick(now, msg).await?;
+        self.callback_senders.extend(callback_senders);
```
```suggestion
        self.callback_senders.append(&mut callback_senders);
```

yellowstone-grpc-tools/src/scylladb/agent.rs
```diff
@@ -0,0 +1,331 @@
...
+
+        AgentHandler {
+            name: agent_name,
+            sender,
+            abort_handle: Arc::new(abort_handle),
+            deadletter_queue: None,
+        }
+    }
+
+    pub fn until_one_agent_dies(&mut self) -> impl Future<Output = anyhow::Result<Nothing>> + '_ {
```
fanatid: why return `impl Future` instead of `async fn`?

.github/workflows/test.yml
lvboudre: Since I overwrote the github commit history it was originally related to juan changes, I'll try to use action-rs back again, I don't remember why we went for dtolnay

yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
lvboudre: oops, leftover from previous debugging

yellowstone-grpc-tools/src/scylladb/types.rs
lvboudre: Oh nice I was not aware we could collect an iterator of result into a result of vector.
I was looking for this on the rust Iterator documentation, there was try_collect which is not stable yet.
Nice to know.

@fanatid However since I checked before that is_err() was false before unwrap should it be the same as `try_into()?`?
If there is an error and we try unwrap/try_into()? it will panic either way.

yellowstone-grpc-tools/src/scylladb/agent.rs
lvboudre: Nice I was not aware of `bail!` macro.

yellowstone-grpc-tools/src/scylladb/agent.rs
lvboudre: since until_one_agent_dies is used inside tokio::select! I need to make sure until_one_Agent_dies returns a future that is cancel safe. This is the case of `join_next`.
I am not sure of how the cancel semantic is propagated in nested async/await futures.

yellowstone-grpc-tools/src/scylladb/agent.rs
lvboudre: It won't work since I `result` is `Result<Result<...>>` which I need to unnest to fit with my return type

Juanito87:
action-rs/toolchain was changed to dtolnay/rust-toolchain because action-rs project has been archived. 
last release is from 2020, throw warnings regarding node.js on each run. @lvboudre @fanatid

yellowstone-grpc-tools/src/scylladb/agent.rs
fanatid: ah, right
maybe in the future... https://doc.rust-lang.org/std/result/enum.Result.html#method.flatten or `result.map_err(Into::into).and_then(|x| x)`

fanatid (approved):

PR #336, opened 2024-05-14
tools: added producer lock + refactor period commitment

lvboudre:


yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -355,54 +344,80 @@ pub enum ScyllaSinkError {
...
-        );
-        Ok(Some(rows))
+            AND shard_id = ?
+            and period = ?
+        ORDER BY offset desc
+        PER PARTITION LIMIT 1        
+    "###;
+    let max_offset_for_shard_period_ps = session.prepare(query_max_offset_for_shard_period).await?;
+
+    let mut js: JoinSet<anyhow::Result<(i16, i64)>> = JoinSet::new();
```
fanatid: probably we can use `futures::future::try_join_all` here?

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -476,36 +617,62 @@ impl ScyllaSink {
...
     }
 
+    pub async fn shutdown(self) -> anyhow::Result<()> {
+        warn!("Shutthing down scylla sink...");
+        self.router_handle.send(ClientCommand::Shutdown).await?;
+        for (i, ss) in self.shard_handles.into_iter().enumerate() {
+            let result = ss.await;
+            if result.is_err() {
+                error!("shard {} error: {:?}", i, result);
+            }
```
```suggestion
            if let Err(error) = ss.await {
                error!("shard {i} error: {error:?}");
            }
```
should be more readable than `shard 0 error: Err(error_here)` (we do not need `Err()` :thinking: )

fanatid (approved):

yellowstone-grpc-tools/src/scylladb/sink.rs
lvboudre: I was exactly looking for something like this but JoinSet didn't have `join_all` function. thanks.

PR #340, opened 2024-05-16
added max seen slot on the sink + added producer healthcheck

lvboudre:


yellowstone-grpc-tools/src/bin/grpc-scylladb.rs
```diff
@@ -179,7 +166,14 @@ impl ArgsAction {
...
+            .transpose();
+
+            if message.is_err() {
+                error!("geyser plugin disconnected: {:?}", message);
+                break;
+            }
+
+            let message = message?;
 
             if let Some(message) = message {
```
fanatid:
```suggestion
            if let Err(error) = &message {
                error!("geyser plugin disconnected: {error:?}");
                break;
            }

            if let Some(message) = message? {
```

yellowstone-grpc-tools/src/scylladb/sink.rs
```diff
@@ -256,26 +257,16 @@ impl Shard {
...
+                    self.session
+                        .execute(&commit_period_ps, (producer_id, shard_id, curr_period - 1))
+                        .await?;
+                    info!(
+                        "shard={},producer_id={:?} committed period: {}: time to commit: {:?}",
+                        shard_id,
+                        self.producer_id,
+                        curr_period,
+                        t.elapsed()
+                    );
```
```suggestion
                    info!(
                        shard = shard_id,
                        producer_id = ?self.producer_id,
                        committed_period = curr_period,
                        time_to_commit = ?t.elapsed()
                    );
```

with tracing it's possible to use fields, the benefit of this is that if the output is json then getting field values is much easier
no need to change, just an example of tracing feature

fanatid (approved):

yellowstone-grpc-tools/src/scylladb/sink.rs
lvboudre: much nicer indeed!

PR #369, opened 2024-07-02
proto: add lamports filter

### LINKED ISSUE #275: add lamports filter to accounts
with lamports filter it would be possible to receive notifications about deleted accounts (owner = system program + lamports = 0)

fanatid:
Closes #275 

@grooviegermanikus @vovkman what do you think about these proto changes? with filter by lamports it would be possible to subscribe on removed accounts


fanatid:
@grooviegermanikus @vovkman can you please review the changes?

PR #380, opened 2024-07-23
proto: fix go_package location

JordanSinko:


JordanSinko:
One of the proto files `go_package` paths were updated and not the other. Unable to refresh using protoc without updating both.

https://github.com/rpcpool/yellowstone-grpc/pull/376

fanatid:
Thank you!

PR #389, opened 2024-08-05
feat: support zstd for grpc

ZhengYuTay:
- jupiter-swap-api have been on gzip compression and it's been great at handling some of the load. Would love to try out zstd as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,11 +1,11 @@
 [workspace]
 resolver = "2"
 members = [
-    "examples/rust", # 1.14.0+solana.2.0.2
+    "examples/rust",           # 1.14.0+solana.2.0.2
     "yellowstone-grpc-client", # 1.16.0+solana.2.0.2
     "yellowstone-grpc-geyser", # 1.16.0+solana.2.0.2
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -75,3 +75,8 @@ yellowstone-grpc-proto = { path = "yellowstone-grpc-proto", version = "=1.15.0+s
 [profile.release]
 lto = true
 codegen-units = 1
+
+# https://github.com/anza-xyz/agave/blob/70efe819bcf70476926f06c473f29015b3269536/Cargo.toml#L549C1-L551C49
+[patch.crates-io.curve25519-dalek]
+git = "https://github.com/anza-xyz/curve25519-dalek.git"
+rev = "b500cdc2a920cd5bff9e2dd974d7b97349d61464"
```
ZhengYuTay, line 82: this have some depedency conflict when building, adding this patch that is same as agave's version to resolve it.

fanatid:
Added in https://github.com/rpcpool/yellowstone-grpc/pull/391, thank you!

ZhengYuTay:
thanks!

PR #406, opened 2024-08-22
Remove deadlock on snapshot subscribing

kespinola:
### Issue

The initial while loop never breaks so the connection which claims the snapshot channel never starts receiving messages. 

## Fix
Allow filter update and pushing of snapshots messages to happen at the same time.


fanatid:
correct pr: #408

PR #470, opened 2024-11-24
Force client disconnects if the Geyser gRPC is lagging behind 

pmantica11:


linuskendall:
Hi @pmantica11 

Thanks for the contribution!

I have a question about the PR - this seems some level of complexity to add to the plugin that would better live outside of the plugin or is there a good reason to manage it within the plugin?

PR #477, opened 2024-11-30
proto: add `from_slot`

fanatid:


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -808,6 +875,57 @@ impl GrpcService {
...
+                                            let _ = stream_tx.send(Err(Status::internal("failed to get replay response"))).await;
+                                        });
+                                        break 'outer;
+                                    };
+
+                                    let Some(mut messages) = messages else {
+                                        info!("client #{id}: replay from {replay_from_slot} is not available");
+                                        tokio::spawn(async move {
+                                            let _ = stream_tx.send(Err(
+                                                Status::internal(format!("replay from {replay_from_slot} is not available"))
```
linuskendall: could we inform the client about what is the "oldest" slot? maybe let the client pick if you want to replay from that older slot in case the slot they requested isn't available?

linuskendall (approved):

yellowstone-grpc-geyser/src/grpc.rs
fanatid: Yeah, I though about it, should be good to provide such info but problem is that we constantly remove messages so on next request client can receive error for returned slot :slightly_smiling_face:  anyway, I'll add it to message and client will be able to request `returned_slot + n`

xiaying-peng:
is from_slot working?

fanatid:
Yes, `from_slot` works if plugin properly configured

suniljalandhra:
Can ```from_slot``` works for any arbitrary in history ?

fanatid:
> Can `from_slot` works for any arbitrary in history ?

No, it can't

PR #481, opened 2024-12-01
geyser: add worker_threads and affinity

fanatid:


andreisilviudragnea:
This PR made the project unbuildable on macOS, because [affinity](https://crates.io/crates/affinity) crate does not support macOS.

fanatid:
is it possible to run test-validator on macOS? :thinking:

andrei-pump:
@fanatid Yes, and I have been using it for a long time now. If you can do something to fix the master branch, it would be amazing, as for now I have to use https://github.com/rpcpool/yellowstone-grpc/releases/tag/v3.0.1+solana.2.0.18, which is the latest tag which still works on macOS.

andreisilviudragnea:
@fanatid Yes, and I have been using it for a long time now. If you can do something to fix the master branch, it would be amazing, as for now I have to use https://github.com/rpcpool/yellowstone-grpc/releases/tag/v3.0.1+solana.2.0.18, which is the latest tag which still works on macOS.

PR #497, opened 2024-12-16
geyser: Add option to allow the injection of a root authority

c410-f3r:
The configuration file of the geyser plugin allows the addition of certificates but it is not currently possible to add a root authority in the Rust client.

https://github.com/rpcpool/yellowstone-grpc/blob/3192b5508d91b342a7e2a714df088efc72598558/yellowstone-grpc-geyser/config.json#L12-L15

For example, with the newly introduced parameter it is now possible to interact with self-made CAs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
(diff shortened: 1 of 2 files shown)


fanatid:
Thank you! Can you also please add a note to CHANGELOG.md?

PR #508, opened 2025-01-06
geyser: add connection slot lag metric

fanatid:


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1089,7 +1106,7 @@ impl GrpcService {
 
 #[tonic::async_trait]
 impl Geyser for GrpcService {
-    type SubscribeStream = ReceiverStream<TonicResult<FilteredUpdate>>;
+    type SubscribeStream = ReceiverStream;
```
lvboudre, line 1109: We should use `BoxStream<'static, TonicResult<FilteredUpdated>>` instead

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1297,3 +1317,44 @@ impl Geyser for GrpcService {
...
+            {
+                if slot > self.max_slot {
+                    self.max_slot = slot;
+                    metrics::connections_slot_lag_observe(slot);
+                }
+            }
+        }
+        Poll::Ready(value)
+    }
+}
```
lvboudre, line 1360: We should use boxed the tokio stream and use `map` instead of implementing our own `Stream` trait.

lvboudre (changes requested):
Conflicting files.

PR #509, opened 2025-01-07
add support for utilizing npm package in ESM environments

solpkr1:
- use --ts_proto_opt=esModuleInterop=true arg when generating with proto
- split esm / cjs into 2 seperate tsconfigs
- add npm run command to copy compiled encoding folder contents to esm/cjs
- add script to fix the imports in the ESM version

Tested with CJS / ESM / TS enviroments for Subsrbribing to transactions for a program account

"esModuleInterop": true is required in projects tsconfig.json compilerOptions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,21 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Fixes
 
-- nodejs: fix port for https ([#502](https://github.com/rpcpool/yellowstone-grpc/pull/502))
-
 ### Features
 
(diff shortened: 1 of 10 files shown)


WilfredAlmeida:
1. I've tested unary calls, accounts & transactions subscriptions and WASM parsing via the [TypeScript example](https://github.com/AphelionTrading/yellowstone-grpc/tree/esm-package-support/examples/typescript), it all works fine.
2. Users might need to set `"esModuleInterop": true` in their `tsconfig.json` if not already, otherwise they'll face errors

PR #511, opened 2025-01-13
Expose TLS Config as part of the client

pmantica11:
I am getting:
```
Error: tonic::transport::Error(Transport, ConnectError(HttpsUriWithoutTlsSupport(())))
```

When using the latest gRPC client against an https endpoint. To fix this I need to specify a ClientTLSConfig, but I cannot because it's not exposed as part of the yellowstone-grpc-client. 

Relevant issue:
https://github.com/hyperium/tonic/issues/1817

Note:
For some reason, I do not get this error when using the 1.15 client. Only 2.0.0 and above.


fanatid:
https://github.com/rpcpool/yellowstone-grpc/commit/f6adda0ca785a361f4afaba55992d74a4838cdd3

PR #519, opened 2025-01-29
Force gRPC server panics to log

pmantica11:
Panics won't log necessarily without this.


linuskendall (approved):
Looks good and really sorry for the merge delay on this one.

Issue #523, opened 2025-02-05
FIRST_SHRED_RECEIVED doesn't work with Agave 2.1.11


thiras:
As mentioned #487 , we have our node running agave 2.1.11 with `v4.2.2+solana.2.1.11`. We don't receive any information when the CommitmentLevel set to FIRST_SHRED_RECEIVED

Our entrypoint;

```bash
#!/bin/bash
exec agave-validator \
    --identity /home/sol/validator-keypair.json \
    --known-validator 7Np41oeYqPefeNQEHSv1UDhYrehxin3NStELsSKCT4K2 \
    --known-validator GdnSyH3YtwcxFvQrVVJMm1JhTS4QVX7MFsX56uJLUfiZ \
    --known-validator DE1bawNcRJB9rVm3buyMVfr8mBEoyyu73NBovf2oXJsJ \
    --known-validator CakcnaRDHka2gXyfbEd2d3xsvkJkqsLw2akB3zsN1D2S \
    --geyser-plugin-config /home/sol/geyser-config.json \
    --full-rpc-api \
    --no-voting \
    --ledger /home/sol/ledger \
    --accounts /home/sol/accounts \
    --log /home/sol/solana-rpc.log \
    --rpc-port 8899 \
    --rpc-bind-address 0.0.0.0 \
    --private-rpc \
    --dynamic-port-range 8000-8020 \
    --entrypoint entrypoint.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint2.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint3.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint4.mainnet-beta.solana.com:8001 \
    --entrypoint entrypoint5.mainnet-beta.solana.com:8001 \
    --expected-genesis-hash 5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d \
    --wal-recovery-mode skip_any_corrupted_record \
    --limit-ledger-size
```

```json
{
  "libpath": "yellowstone-grpc/target/release/libyellowstone_grpc_geyser.so",
  "log": {
    "level": "debug"
  },
  "tokio": {
    "worker_threads": 16,
    "affinity": "48-63"
  },
  "grpc": {
    "address": "0.0.0.0:10000",
    "compression": {
      "accept": ["gzip", "zstd"],
      "send": ["gzip", "zstd"]
    },
    "server_http2_adaptive_window": null,
[...]

PR #525, opened 2025-02-05
New way to specify slot commitment levels

linuskendall:
Move way to slot commitment levels 

The current way of expanding commitment levels is confusing see #523

### DIFF
--- a/yellowstone-grpc-proto/proto/geyser.proto
+++ b/yellowstone-grpc-proto/proto/geyser.proto
@@ -21,6 +21,12 @@ enum CommitmentLevel {
   PROCESSED = 0;
   CONFIRMED = 1;
   FINALIZED = 2;
+}
+
+enum SlotCommitmentLevel {
+  PROCESSED = 0;
+  CONFIRMED = 1;
+  FINALIZED = 2;
   FIRST_SHRED_RECEIVED = 3;
   COMPLETED = 4;
   CREATED_BANK = 5;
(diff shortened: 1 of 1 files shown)


yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -21,6 +21,12 @@ enum CommitmentLevel {
   PROCESSED = 0;
   CONFIRMED = 1;
   FINALIZED = 2;
+}
+
+enum SlotCommitmentLevel {
+  PROCESSED = 0;
+  CONFIRMED = 1;
+  FINALIZED = 2;
```
fanatid, line 29:
```
  geyser.proto:27:3: "PROCESSED" is already defined in "geyser".
  geyser.proto:27:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "PROCESSED" must be unique within "geyser", not just within "SlotStatus".
  geyser.proto:28:3: "CONFIRMED" is already defined in "geyser".
  geyser.proto:28:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "CONFIRMED" must be unique within "geyser", not just within "SlotStatus".
  geyser.proto:29:3: "FINALIZED" is already defined in "geyser".
  geyser.proto:29:3: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "FINALIZED" must be unique within "geyser", not just within "SlotStatus".
```
:face_with_diagonal_mouth:

fanatid:
Moved to #529

PR #536, opened 2025-02-10
[DO NOT MERGE] Branch for running yellowstone grpc in macos

kewei5zhang:


fanatid:
is this required for development purposes? I'd be happy to merge if you leave only code with cfg and add condition by os in dependencies

PR #558, opened 2025-03-18
Added metrics + remove sleep overhead in geyser_loop

lvboudre:
@linuskendall @fanatid 
I added comment in the PR to explain the changes.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -12,12 +12,10 @@ use {
         pubkey::Pubkey,
     },
     std::{
-        collections::{BTreeMap, HashMap},
-        sync::{
+        collections::{BTreeMap, HashMap}, str::FromStr, sync::{
             atomic::{AtomicUsize, Ordering},
             Arc,
-        },
-        time::SystemTime,
+        }, time::SystemTime
     },
     tokio::{
         fs,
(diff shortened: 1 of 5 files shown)


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -510,21 +508,24 @@ impl GrpcService {
...
+        // If we assume 3000 TPS, 31 messages will be processed in ~10ms
         const PROCESSED_MESSAGES_MAX: usize = 31;
-        const PROCESSED_MESSAGES_SLEEP: Duration = Duration::from_millis(10);
 
         let mut msgid_gen = MessageId::default();
         let mut messages: BTreeMap<u64, SlotMessages> = Default::default();
         let mut processed_messages = Vec::with_capacity(PROCESSED_MESSAGES_MAX);
         let mut processed_first_slot = None;
-        let processed_sleep = sleep(PROCESSED_MESSAGES_SLEEP);
-        tokio::pin!(processed_sleep);
```
lvboudre, line 521: I completely removed `processed_sleep` as `Sleep` object acquires `Mutex` on Cancel and Reset which is done on each loop iteration.
This should remove slight overhead in the main loop of Geyser plugin.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -510,21 +508,24 @@ impl GrpcService {
...
         let (_tx, rx) = mpsc::channel(1);
         let mut replay_stored_slots_rx = replay_stored_slots_rx.unwrap_or(rx);
 
         loop {
+            let t = Instant::now();
             tokio::select! {
-                Some(message) = messages_rx.recv() => {
+                maybe = messages_rx.recv() => {
```
lvboudre, line 527: I remove the `Some(message)` for `maybe` since the `select!` used to look like this:

```rust
tokio::select! {
  Some(..) = message_rx.recv() => { ... }
  Some(...) = replay_from_slot_rx.recv() => { ... }
  else => { break }
}
```

If we disable replay from slot feature in the plugin, then `replay_from_slot_rx` is a channel that never returns, but never drop. So it will never return `None`.

Because of this, if the `message_rx` was to close, this could create a deadlock.

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -67,6 +67,18 @@ lazy_static::lazy_static! {
         Opts::new("missed_status_message_total", "Number of missed messages by commitment"),
         &["status"]
     ).unwrap();
+
+
+    static ref GEYSER_LOOP_ITERATION_DURATION: Histogram = Histogram::with_opts(
+        HistogramOpts::new("geyser_loop_iteration_duration_microsecond", "Duration of ONE geyser loop iteration in MICROSECOND")
+            .buckets(vec![100.0, 250.0, 500.0, 1000.0, 2000.0, 5000.0, 10000.0])
+        ).unwrap();
```
lvboudre, line 79: This metric measure the time it took to do one loop iterateion in `geyser_loop`.
The unit is in microsecond, not millisecond.
If we want to do 8k message/second, each loop iteration should not be greater than 100 microseconds.

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -67,6 +67,18 @@ lazy_static::lazy_static! {
...
+
+    static ref FILTER_UPDATE_DURATION_VEC: HistogramVec = HistogramVec::new(
+        HistogramOpts::new("filter_update_duration_ms", "Filter Duration Observations per geyser message in MILLISECOND")
+            .buckets(vec![1.0, 2.0, 3.0, 5.0, 8.0, 13.0, 21.0, 34.0, 55.0, 89.0, 144.0]),
+        &["x_subscription_id"]
+    ).unwrap();
```
lvboudre, line 85: I added this metrics to measure how much time is spent to filter each message send to each client.
Since the filter code is quite complex and that we have latency issue, I want to confirm if the filter implementations cause too much overhead.

PR #560, opened 2025-03-20
Experimental perf

lvboudre:
I removed all `timestamp` field from each `Message` variants.
Why? Because `Message` enum used to weight 72 bytes and the timestamp fields alone was 16 bytes.

When `Message` weights `72` bytes, the message cannot fit into one cache-line.
By removing the timestamp field, the `Message` enum size becomes `56` bytes.

This is really important because modern CPU memory fetching works in cache-line granularity. By making sure `Message` is `<= 64 bytes` we make sure that each `Message` does not need additional round-trip to memory.

Inside tight loops, when CPU pre-fetching kicks in, we will benefits from the fact that Latest Skylake Intel and AMD CPU can pre-fetch 128 bytes at a time (two cache-line).

Instead of using tokio's `mspc::UnboundedSender` in between the agave world and the `PluginInner`, I switched to a crossbeam.

Why?

Tokio's channel is fast but is implemented using linked-list of fixed size "arrays" of length `32` called `Block`.  
Here's a [link to tokio block capacity](https://github.com/tokio-rs/tokio/blob/0ec4d0db4d24af74cea40cf0ceed6b7a0a5ff183/tokio/src/sync/mpsc/mod.rs#L138)

The `Block` data is contiguous in-memory, not the linked list of `Block`.
Linked-list data structure creates higher chances of "cache-miss" inside the CPU.

Unlike Tokio, crossbeam `bounded` channel is a fixed array size of contiguous memory: minimising cache miss and maximising  OS page utilization.

Crossbeam `bounded` channel is an array of `Slot`, each slot is :
```rust
pub struct Slot<T> {
  header: AtomicUsize,
  data: UnsafeCell<MaybeUninit<T>>
}
```
the `data` field size is the same as `T` and has the same alignment as `T`.
As for `AtomicUsize`, its size is `8` bytes, making `Slot<Message>` weigh `64 bytes` total.
Perfectly fit in one cache-line exactly.

Since `Slot` are contiguous in physical memory, inside tight-loop we will be able to leverage `cache-prefetching` that is available in all modern CPU.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -6,7 +6,6 @@ use {
     },
     anyhow::Context,
     log::{error, info},
-    prost_types::Timestamp,
     solana_sdk::{
         clock::{Slot, MAX_RECENT_BLOCKHASHES},
         pubkey::Pubkey,
@@ -18,11 +17,9 @@ use {
             atomic::{AtomicUsize, Ordering},
             Arc,
         },
-        time::SystemTime,
     },
     tokio::{
         fs,
-        runtime::Builder,
         sync::{broadcast, mpsc, oneshot, Mutex, Notify, RwLock, Semaphore},
         task::spawn_blocking,
         time::{sleep, Duration, Instant},
@@ -324,6 +321,7 @@ impl SlotMessages {
 
 type BroadcastedMessage = (CommitmentLevel, Arc<Vec<(u64, Message)>>);
 
+#[allow(dead_code)]
 enum ReplayedResponse {
     Messages(Vec<(u64, Message)>),
     Lagged(Slot),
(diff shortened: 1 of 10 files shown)


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -504,339 +508,356 @@ impl GrpcService {
...
+                        match e {
+                            crossbeam_channel::TryRecvError::Empty => continue,
+                            crossbeam_channel::TryRecvError::Disconnected => {
+                                log::error!("geyser_rx disconnected");
+                                break 'outer;
+                            }
                         }
                     }
+                }
+            }
```
lvboudre: `geyser_loop` is not async anymore, removing any Future scheduling overhead.
Using `try_recv` I avoid any thread parking that may put the thread to sleep and make the thread context switch.
We now wait for 64 messages before processing them, this is because 64 crossbeam channel "entry/slot" will fit inside a memory page minimizing page faulting.

yellowstone-grpc-proto/src/plugin/filter/message.rs
```diff
@@ -105,24 +105,16 @@ impl prost::Message for FilteredUpdate {
...
-        filters: FilteredUpdateFilters,
-        message: FilteredUpdateOneof,
-        created_at: Timestamp,
-    ) -> Self {
+    pub fn new(filters: FilteredUpdateFilters, message: FilteredUpdateOneof) -> Self {
         Self {
             filters,
             message,
-            created_at,
+            created_at: Timestamp::from(SystemTime::now()),
```
lvboudre, line 112: I only put timestamp when we create the "gRPC" message to send to the our client. It is less precise than putting the timestamp right when we first got a hold of the new geyser event, but it prevents geyser from doing bad cache performance in tight loop.

leafaar (approved):
very good!

lvboudre:
won't do/don't have time.

PR #562, opened 2025-04-02
Remove solana-sdk

kevinheavey:
This removes solana-sdk as a direct dependency and replaces it with the smaller component crates for the sake of having lighter dependencies.

It also bumps spl-token-2022 to v8, which removes an indirect solana-program dep.

There are still some deps pulling in solana-program, but solana-sdk is gone


kespinola (approved):

WilfredAlmeida:
Closing this in favour of https://github.com/rpcpool/yellowstone-grpc/pull/590

Thanks @kevinheavey for the PR

PR #574, opened 2025-04-22
fix: use discrete return type instead of impl

OliverNChalk:
The use of `impl Interceptor` in the return type for these functions prevents downstream consumers from storing the client in a struct (without the use of `dyn`). This is because `impl Interceptor` cannot be named. Given these functions actually return `InterceptorXToken` we seem to just be destroying type information for no reason?

### DIFF
--- a/yellowstone-grpc-client/src/lib.rs
+++ b/yellowstone-grpc-client/src/lib.rs
@@ -1,11 +1,7 @@
 pub use tonic::{service::Interceptor, transport::ClientTlsConfig};
 use {
     bytes::Bytes,
-    futures::{
-        channel::mpsc,
-        sink::{Sink, SinkExt},
-        stream::Stream,
-    },
+    futures::{channel::mpsc, sink::SinkExt},
     std::time::Duration,
     tonic::{
         codec::{CompressionEncoding, Streaming},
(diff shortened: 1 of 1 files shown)


kespinola (approved):
LG2M.

The struct can also be made generic if something like mocking requires working with a trait instead of a struct.


```rust
struct MyClient<S, St> {
    sink: S,
    stream: St,
}

impl<S, St> MyClient<S, St>
where
    S: Sink<SubscribeRequest, Error = mpsc::SendError>,
    St: Stream<Item = Result<SubscribeUpdate, Status>>,
{
    fn do_something(&mut self) { ... }
}
```

lvboudre:
I am a bit concern.

The Streaming<SubscribeUpdate> is ok,
but going from impl Sink to UnboundedSender prevent us from changing this bounded sender.
Moreover, this will break current client implementation because Sink is async API all-the-way:
the `send` is async in Sink, while the concrete type `UnboundedSender::send` is not-async

yellowstone-grpc-client/src/lib.rs
```diff
@@ -102,8 +96,8 @@ impl<F: Interceptor> GeyserGrpcClient<F> {
     pub async fn subscribe(
         &mut self,
     ) -> GeyserGrpcClientResult<(
-        impl Sink<SubscribeRequest, Error = mpsc::SendError>,
-        impl Stream<Item = Result<SubscribeUpdate, Status>>,
+        mpsc::UnboundedSender<SubscribeRequest>,
```
lvboudre, line 99: The Streaming is ok,
but going from impl Sink to UnboundedSender prevent us from changing this to another of channel in future say bounded Sender.

However, my main issue with this proposal is : itwill break current client implementation because Sink is async API all-the-way:
the method `Sink::send` from `SinkExt` is `async`, while the concrete type `UnboundedSender::send` is not `async`.

Can we use add `+ Clone` to trait bound instead for the Sink? 
By doing so it will not break any existing client code, but allow you to Clone it.
wdyt?

lvboudre (changes requested):

PR #580, opened 2025-05-02
feat: add metrics.rs, affinity_linux, end to end latency metrics

vovkman:
Adds

1. `affinity-linux` so that builds work on mac

2. `metrics.rs` 
  - so that prometheus isn't the only supported source of metrics
  - adds feature flag for statsd

3. `message_send_latency_ms` metric to tell time from message creation to transmission


yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -35,33 +41,17 @@ lazy_static::lazy_static! {
         &["buildts", "git", "package", "proto", "rustc", "solana", "version"]
     ).unwrap();
 
-    static ref SLOT_STATUS: IntGaugeVec = IntGaugeVec::new(
-        Opts::new("slot_status", "Lastest received slot from Geyser"),
-        &["status"]
-    ).unwrap();
+    static ref SLOT_STATUS: Arc<AtomicI64> = Arc::new(AtomicI64::new(0));
```
lvboudre, line 44: Atomic64 cannot replace a IntGaugeVec (vectorized metric).

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -192,13 +182,6 @@ impl PrometheusService {
                 };
             }
             register!(VERSION);
-            register!(SLOT_STATUS);
-            register!(SLOT_STATUS_PLUGIN);
-            register!(INVALID_FULL_BLOCKS);
-            register!(MESSAGE_QUEUE_SIZE);
-            register!(CONNECTIONS_TOTAL);
-            register!(SUBSCRIPTIONS_TOTAL);
-            register!(MISSED_STATUS_MESSAGE);
```
lvboudre, line 201: This will break the Prometheus Service.

lvboudre (changes requested):
Two things:
- This PR does not preserve existing prometheus metrics and `/debug_clients` endpoint.
- This PR needs to be rebase off master branch.

I am fine with unifying metrics, though we need to make sure it migrate legacy prometheus feature as-is so current user of the plugin does not lose features.

lvboudre:
inactivity

Issue #584, opened 2025-05-12
Opaque `Interceptor`  in `GrpcGeyserClient` prevents cloning of client


jmininger:
Currently the only way to return a `GeyserGrpcClient` are connect` and connect_lazy` methods described below:
```
    pub async fn connect(self) -> GeyserGrpcBuilderResult<GeyserGrpcClient<impl Interceptor>> {
        let channel = self.endpoint.connect().await?;
        self.build(channel)
    }

    pub fn connect_lazy(self) -> GeyserGrpcBuilderResult<GeyserGrpcClient<impl Interceptor>> {
        let channel = self.endpoint.connect_lazy();
        self.build(channel)
    }
```
These both return `GeyserGrpcClient<impl Interceptor>`.  This opaque type prevents the client from:
1) Being stored in a struct and
2) Prevents the addition of a `Clone` on `GeyserGrpcClient` which would allow the client to reuse the same underlying connection. 

If you look at the private `build` method that the two constructors use, you can see that they both just create an `InterceptorToken`

```
        let interceptor = InterceptorXToken {
            x_token: self.x_token,
            x_request_snapshot: self.x_request_snapshot,
        };

```

It would be great to be able to cheaply clone the GeyserGrpcClient around

jmininger:
It looks there is a PR up: #574 but would also be great to add a `Clone` in addition

PR #589, opened 2025-05-27
Add Clone support and explicit types to GeyserGrpcClient

jmininger:
### Summary 
This PR adds `Clone` capability to `GeyserGrpcClient` and improves type clarity by using explicit types instead of opaque `impl Interceptor` types in client constructors. It addresses #584 and also has some overlap with #574

### Opaque Type Limitations
The previous use of impl Interceptor as an opaque return type prevented the client from being stored in structs that need explicit type parameters. When a function returns GeyserGrpcClient<impl Interceptor>, the concrete interceptor type is hidden and unnameable, making it impossible to specify that type in generic contexts.


kespinola (approved):
Nothing seems to merit the impl. InterceptorXToken is simple struct with x_token and x_request_snapshot which can be stubbed easy enough in specs (even though there isn't any).

lvboudre (approved):

jmininger:
@lvboudre just checking to see if you still intend on merging this ?

lvboudre:
> @lvboudre just checking to see if you still intend on merging this ?

Yes, but I might want to consolidate multiple PRs.

PR #609, opened 2025-08-01
Better load monitoring

lvboudre:
Added sevral metrics for geyser and load tracking client session.
Client load uses EMA (Exponential-Moving-Average) to smooth out metrics and remove noise and extreme short-live spike in prometheus/grafana.


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -980,11 +1042,17 @@ impl GrpcService {
...
+
                         if commitment == filter.get_commitment_level() {
                             for (_msgid, message) in messages.iter() {
                                 for message in filter.get_updates(message, Some(commitment)) {
+                                    let proto_size = message.encoded_len().min(u32::MAX as usize) as u32;
                                     match stream_tx.try_send(Ok(message)) {
-                                        Ok(()) => {}
+                                        Ok(()) => {
+                                            metrics::incr_grpc_message_sent_counter(&subscriber_id);
+                                            metrics::incr_grpc_bytes_sent(&subscriber_id, proto_size);
```
lvboudre, line 1043: @linuskendall 
This is the main sending code to the client.
Here I measure the size of the proto payload I will send to the client via gRPC.
Only payload that passes client's filter will be measure.
Finnally, I increase the subscriber message counter.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -878,8 +916,28 @@ impl GrpcService {
...
+
+                set_subscriber_send_bandwidth_load(
+                    &subscriber_id,
+                    stream_tx.estimated_send_rate().per_second() as i64,
+                );
+
+                set_subscriber_recv_bandwidth_load(
+                    &subscriber_id,
+                    stream_tx.estimated_consuming_rate().per_second() as i64,
+                );
```
lvboudre: Here I measure three things:

The overall loop processing pace `set_subscriber_pace` where the unit is "geyser event"/second.

The actual bandwidth load we are sending to the downstream client (only filtered data matching client's filters).
The actual bandwidth consumption rate a client do per second.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1142,6 +1222,12 @@ impl Geyser for GrpcService {
             .and_then(|h| h.to_str().ok().map(|s| s.to_string()))
             .unwrap_or_else(|| "".to_owned());
 
+        let subscriber_id = request
+            .metadata()
+            .get("x-subscription-id")
+            .and_then(|h| h.to_str().ok().map(|s| s.to_string()))
+            .or(request.remote_addr().map(|addr| addr.to_string()));
+
```
lvboudre, line 1222: In order to identitfy a downstream client I check if `x-subscription-id` is present otherwise I use its ip address.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -650,6 +686,7 @@ impl GrpcService {
                         }
                         // Dedup accounts by max write_version
                         Message::Account(msg) => {
+                            metrics::observe_geyser_account_update_received(msg.account.data.len());
```
lvboudre, line 689: This line of code is inside the "geyser loop" which process every geyser event from agave.
I measure the account data size and put it inside an histogram.

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -67,6 +69,51 @@ lazy_static::lazy_static! {
...
+    ).unwrap();
+
+    static ref GEYSER_ACCOUNT_UPDATE_RECEIVED: Histogram = Histogram::with_opts(
+        HistogramOpts::new(
+            "geyser_account_update_data_size_kib",
+            "Histogram of all account update data (kib) received from Geyser plugin"
+        )
+        .buckets(vec![5.0, 10.0, 20.0, 30.0, 50.0, 100.0, 200.0, 300.0, 500.0, 1000.0, 2000.0, 3000.0, 5000.0, 10000.0])
+    ).unwrap();
 }
```
lvboudre, line 117: I opted for an histogram to measure account data our geyser plugin receives since histogram gives us two other metrics for "free":

the total sum of data (`geyser_account_update_data_size_kib_sum`) and the account update counts (`geyser_account_update_data_size_kib_count`).

The bucket is based of previous work I did for fumarole.
The P90 of account data size should be below 5KiB.
The P95 should be < 20kIB.
About 1% of account data can be above 1MiB.
The max size (bucket) is ~10mb which match the max account data size.

yellowstone-grpc-geyser/src/util/ema.rs
```diff
@@ -0,0 +1,388 @@
...
+        RwLock,
+    },
+    time::{Duration, Instant},
+};
+
+///
+/// Exponential Moving Average (EMA) for load tracking.
+///
+#[derive(Debug)]
+pub struct Ema {
```
lvboudre, line 16: This is Exponential-Moving-Avg algorithm to compute an average load. 

This is important to use to have a smoothed view of the load we are sending to downstream client and remove noise from graph, thus helping us with better visibility.

.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
   release:
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-22.04]
```
leafaar: Let's also add 24.04 for future proof

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -67,6 +69,59 @@ lazy_static::lazy_static! {
         Opts::new("missed_status_message_total", "Number of missed messages by commitment"),
         &["status"]
     ).unwrap();
+
+    static ref GRPC_MESSAGE_SENT: IntCounterVec = IntCounterVec::new(
```
lvboudre, line 73: Measure how many geyser event we sent to each subscriber.

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -67,6 +69,59 @@ lazy_static::lazy_static! {
...
+        Opts::new(
+            "grpc_message_sent_count",
+            "Number of message sent over grpc to downstream client",
+        ),
+        &["subscriber_id"]
+    ).unwrap();
+
+    static ref GRPC_BYTES_SENT: IntCounterVec = IntCounterVec::new(
```
lvboudre, line 81: Measures how many bytes (protobuffer encoded data) to each subscriber

leafaar (approved):

PR #614, opened 2025-08-18
fix: resolve Windows compilation issues for protobuf-src

xun0n:
Windows users now need to:
1. Manually download protoc binary from official releases
2. Set PROTOC environment variable pointing to the executable


lvboudre:
Windows supports is not a goal of this project.

PR #616, opened 2025-08-20
fix(slot_messages): handle lower write_version by dropping messages

dimitriedavid:
This PR fixes an edge case in account update deduplication for `slot_messages.messages`.

### Background

Yellowstone gRPC saves account updates into `slot_messages.messages`.
The intent is to keep only the latest (highest `write_version`) account update per pubkey within a slot.
Deduplication is tracked via `slot_messages.accounts_dedup`.

### Problem

Previously, when an account update arrived out of order (with a `write_version` **lower** than the highest seen so far), the code failed to discard it.
Because messages are appended to `slot_messages.messages` **before** deduplication logic runs, this resulted in multiple updates being saved for the same pubkey in a slot.

### Fix

Added an else branch to handle the case where the new update’s `write_version` is lower than the latest one.
The new (out-of-order) message is immediately dropped by marking its array entry as `None`.
This ensures only the latest `write_version` per pubkey survives.

### Impact

Prevents saving multiple account updates per pubkey in a slot.
Preserves intended deduplication semantics.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2966,9 +2966,9 @@ checksum = "74233d3b3b2f6d4b006dc19dee745e73e2a6bfb6f93607cd3b02bd5b00797d7c"
 
 [[package]]
 name = "slab"
-version = "0.4.10"
+version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


leafaar (approved):

leafaar:
Thanks for the contribution :grin:

PR #620, opened 2025-09-02
reafactor: upgrade dependencies to agave 3 and fix errors

WilfredAlmeida:


yellowstone-grpc-proto/proto/solana-storage.proto
```diff
@@ -23,6 +23,7 @@ message ConfirmedTransaction {
 message Transaction {
     repeated bytes signatures = 1;
     Message message = 2;
+    string message_hash = 3;
```
linuskendall, line 26: is this change backwards compatible? I can't remember the rules for gRPC backwards compatbility?

leafaar: I believe it is, since it's something new, the old will still work with it

yellowstone-grpc-proto/src/plugin/filter/filter.rs
```diff
@@ -1180,7 +1180,9 @@ mod tests {
             transaction: Arc::new(MessageTransactionInfo {
                 signature: *sig,
                 is_vote: true,
-                transaction: convert_to::create_transaction(&sanitized_transaction),
+                transaction: convert_to::create_transaction(
+                    &sanitized_transaction.to_versioned_transaction(),
+                ),
```
leafaar: Why are you converting the sanitized transaction here if you can just create a versioned one?

yellowstone-grpc-client-nodejs/solana-encoding-wasm/Cargo.toml
```diff
@@ -13,10 +13,11 @@ publish = false
 crate-type = ["cdylib"]
 
 [dependencies]
+getrandom = { version = "0.3", features = ["wasm_js"] }
```
leafaar, line 18: I don't see where we use this dependency? I'm guessing you added to bump because this a dependency of other crate and you needed to update? You can use `cargo update` if that's the case

leafaar (approved):
I added some comments that I'm also sending a commit fixing, they are just nitpicks. Overall very good job, thanks Will!

yellowstone-grpc-client-nodejs/solana-encoding-wasm/Cargo.toml
```diff
@@ -30,4 +30,4 @@ codegen-units = 1
 # wasm32 and spl-token-confidential-transfer-proof-generation
 [patch.crates-io]
 spl-token-2022_6 = { package = "spl-token-2022", version = "6.0.0", git = "https://github.com/rpcpool/token-2022.git", tag = "program@v6.0.0-wasm" }
-spl-token-2022_7 = { package = "spl-token-2022", version = "7.0.0", git = "https://github.com/rpcpool/token-2022.git", tag = "program@v7.0.0-wasm" }
\ No newline at end of file
+spl-token-2022_7 = { package = "spl-token-2022", version = "7.0.0", git = "https://github.com/rpcpool/token-2022.git", tag = "program@v7.0.0-wasm" }
```
lvboudre, line 36: spl-token-2022_7 has dependency on solana crates 2.2.1 it will conflict with `yellowstone-grpc-proto` dependency to version 3, you need to update the private spl-token-2022_7 aswell.

yellowstone-grpc-geyser/src/plugin.rs
```diff
@@ -205,7 +205,10 @@ impl GeyserPlugin for Plugin {
                 ReplicaTransactionInfoVersions::V0_0_1(_info) => {
                     unreachable!("ReplicaAccountInfoVersions::V0_0_1 is not supported")
                 }
-                ReplicaTransactionInfoVersions::V0_0_2(info) => info,
+                ReplicaTransactionInfoVersions::V0_0_2(_info) => {
+                    unreachable!("ReplicaAccountInfoVersions::V0_0_2 is not supported")
+                }
+                ReplicaTransactionInfoVersions::V0_0_3(info) => info,
```
lvboudre, line 211: is it a mistake? We used to support v0_0_2, can't we support both?

lvboudre (changes requested):

yellowstone-grpc-geyser/src/plugin.rs
leafaar: Nope. They changed the format, thus updating the version, see: https://github.com/anza-xyz/agave/commit/6b566ba9cd58a4cb08015744f9db1236ff2f2458

WilfredAlmeida:
Closing this PR in favour of https://github.com/rpcpool/yellowstone-grpc/pull/624

Changes are the same, this PR got caught in git rebase mess due to unsigned commits

PR #624, opened 2025-09-04
feat: agave 3.0 dependency upgrades, ci fixes, version bumps, changelog

WilfredAlmeida:


Lusitaniae (approved):

yellowstone-grpc-client-nodejs/solana-encoding-wasm/Cargo.toml
```diff
@@ -13,10 +13,13 @@ publish = false
 crate-type = ["cdylib"]
 
 [dependencies]
+# getrandom is used by a dependency which does not eble the wasm_js feature and causes compilation for wasm to fail.
```
leafaar:
```suggestion
# getrandom is used by a dependency which does not enable the wasm_js feature and causes compilation for wasm to fail.
```

Cargo.toml
```diff
@@ -20,70 +20,70 @@ keywords = ["solana"]
...
-hyper = "~1.4.1"
-hyper-util = "~0.1.7"
-indicatif = "~0.17.9"
-lazy_static = "1.4.0"
-local-ip-address = "~0.6.1"
-log = "~0.4.17"
-maplit = "1.0.2"
-prometheus = "0.13.2"
+affinity = "^0.1.2"
+agave-geyser-plugin-interface = "^3.0.0"
```
fanatid, line 24: should be `~3.0.0`
you can check list of dependencies here: https://github.com/fanatid/yellowstone-grpc/blob/agave-3.0-deps/Cargo.toml#L22

Buzzec: With Solana's new policy of following semver, `~` should not be required.

fanatid: I also though about it but was not able to find public statement or something like this...  maybe you have some source of it?

PR #625, opened 2025-09-10
plugin: put cancellation token

### LINKED ISSUE #617: [Error] Connectiosn get orphaned on plugin reload
### Disclaimer

Please fill all the required data, if  not, the issue will not be considered and closed without any further review.
Before creating the issue, please do the following:

* Run the config check utility  `cargo run --bin config-check -- --config path_to_your_config/config.json`
* Ensure that solana version and yellowstone-grpc version are running with matching versions
* Ensure that solana and yellowstone-grpc are compiled with the same environment

### Describe your build and run environment 

If build and run environment are different, please add the details for both.

CPU:  9474F
RAM: 768
OS: ubuntu 22.04
Rust version: 1.86
Solana version: 2.3.6
Yellowstone grpc version: v8.0.0+2.3.6

### Config

Please paste your config file here and the result of the config-check:

### Describe your issue

When reloading the plugin after adding `replay_stored_slots`  the existing connections get orphaned and not closed properly.

lvboudre:
Close #617 

# Bug

ot reloading the geyser plugin appears to leave behind "zombie" connections, suggesting that unloading the plugin isn’t properly closing active connections.

On closer inspection of the plugin’s on_unload method, we do call shutdown on the Tokio runtime created during Plugin::load. However, shutdown does not terminate tasks spawned with tokio::spawn, as documented [here](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#shutdown).

> The [shutdown_background](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_background) and [shutdown_timeout](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.shutdown_timeout) methods can be used if waiting forever is undesired. When the timeout is reached, spawned work that did not stop in time and threads running it are leaked.

Because we rely heavily on tokio::spawn for gRPC connections, this behavior likely explains why orphaned connections remain after the plugin is unloaded.

# After 

This PR introduce the usage of [CancellationToken](https://docs.rs/tokio-util/0.7.16/tokio_util/sync/struct.CancellationToken.html) and [TaskTracker](https://docs.rs/tokio-util/0.7.16/tokio_util/task/task_tracker/struct.TaskTracker.html).

All fire-and-forget tokio::spawn calls have been replaced with TaskTracker::spawn, combined with a tree-structured `CancellationToken`. This ensures that every task spawned by the plugin components listens to the same cancellation signal during Plugin::on_unload.

The `CancellationToken` centralizes shutdown logic, allowing us to signal all tasks at once when the plugin is being unloaded. Inside Plugin::on_unload, we then perform a graceful shutdown using `TaskTracker::wait`, ensuring that both auxiliary and gRPC tasks are properly terminated.


yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -150,41 +151,52 @@ impl Drop for DebugClientStatuses {
...
+                            match clients.entry(id) {
+                                HashMapEntry::Occupied(mut entry) => {
+                                    entry.get_mut().filter = filter;
+                                }
+                                HashMapEntry::Vacant(entry) => {
+                                    entry.insert(DebugClientStatus {
+                                        filter,
+                                        processed_slot: 0,
+                                    });
+                                }
```
lvboudre: Here, I added more check in case the value return by `clients_rx` is None to break the loop, the logic is still the same as before.

yellowstone-grpc-geyser/src/plugin.rs
```diff
@@ -32,8 +33,8 @@ pub struct PluginInner {
     snapshot_channel: Mutex<Option<crossbeam_channel::Sender<Box<Message>>>>,
     snapshot_channel_closed: AtomicBool,
     grpc_channel: mpsc::UnboundedSender<Message>,
-    grpc_shutdown: Arc<Notify>,
-    prometheus: PrometheusService,
+    plugin_cancellation_token: CancellationToken,
```
lvboudre, line 36: The Plugin has a single shutdown signal that will propagate throughout the entire system.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -129,75 +130,89 @@ struct BlockMetaStorage {
...
                     }
-                    Message::BlockMeta(msg) => {
-                        storage.blocks.insert(msg.slot, msg);
-                    }
-                    msg => {
-                        error!("invalid message in BlockMetaStorage: {msg:?}");
-                    }
                 }
             }
         });
```
lvboudre: It looks worse than it is: 
I removed the `is_alive` flag to use a `Result<(), ClientSnapshotReplayError>` instead and simply which `while let Some(msg) = rx...` for a `loop` instruction with `tokio::select` and always make sure to `break` when `CancellationToken` is cancel.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -410,8 +436,11 @@ impl GrpcService {
         let (blocks_meta, blocks_meta_tx) = if config.unary_disabled {
             (None, None)
         } else {
-            let (blocks_meta, blocks_meta_tx) =
-                BlockMetaStorage::new(config.unary_concurrency_limit);
+            let (blocks_meta, blocks_meta_tx) = BlockMetaStorage::new(
+                config.unary_concurrency_limit,
+                service_cancellation_token.child_token(),
+                task_tracker.clone(),
```
lvboudre, line 443: Make sure BlockMetaStraoge inherits from GrpcService's `CancellationToken` + `TaskTracker`

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -893,9 +923,19 @@ impl GrpcService {
         mut messages_rx: broadcast::Receiver<BroadcastedMessage>,
         replay_stored_slots_tx: Option<mpsc::Sender<ReplayStoredSlotsRequest>>,
         debug_client_tx: Option<mpsc::UnboundedSender<DebugClientMessage>>,
-        drop_client: impl FnOnce(),
+        cancellation_token: CancellationToken,
+        task_tracker: TaskTracker,
```
lvboudre, line 920: Make sure `client_loop` function inherits from GrpcService's `CancellationToken` + `TaskTracker`

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -905,169 +945,189 @@ impl GrpcService {
...
-            Self::client_loop_snapshot(
+            let result = Self::client_loop_snapshot(
                 id,
                 &endpoint,
                 stream_tx.clone(),
                 &mut client_rx,
                 snapshot_rx,
-                &mut is_alive,
                 &mut filter,
+                cancellation_token.clone(),
```
lvboudre, line 950: Propagate the CancellationToken to the snapshot replay if needed

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -905,169 +945,189 @@ impl GrpcService {
...
-                                }
+            set_subscriber_queue_size(&subscriber_id, stream_tx.queue_size());
+
+            tokio::select! {
+                _ = cancellation_token.cancelled() => {
+                    info!("client #{id}: cancelled");
+                    stream_tx.send(Err(Status::unavailable("server is shutting down try again later"))).await.ok();
+                    break 'outer;
+                }
+                mut message = client_rx.recv() => {
```
lvboudre: The `client_loop` main message loop now actively watch if `CancellationToken` gets canceled to break the loop early.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -905,169 +945,189 @@ impl GrpcService {
...
-                        if commitment == CommitmentLevel::Processed && debug_client_tx.is_some() {
-                            for message in messages.iter() {
-                                if let Message::Slot(slot_message) = &message.1 {
-                                    DebugClientMessage::maybe_send(&debug_client_tx, || DebugClientMessage::UpdateSlot { id, slot: slot_message.slot });
-                                }
+                    if commitment == CommitmentLevel::Processed && debug_client_tx.is_some() {
+                        for message in messages.iter() {
+                            if let Message::Slot(slot_message) = &message.1 {
+                                DebugClientMessage::maybe_send(&debug_client_tx, || DebugClientMessage::UpdateSlot { id, slot: slot_message.slot });
                             }
```
lvboudre, line 1126: THIS IS ALL REFORMATTING : NOTHING CHANGED HERE

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1091,38 +1151,51 @@ impl GrpcService {
...
-                            error!("client #{id}: stream closed");
-                            *is_alive = false;
+        loop {
+            tokio::select! {
+                _ = cancellation_token.cancelled() => {
+                    info!("client #{id}: cancelled");
+                    return Err(ClientSnapshotReplayError::Cancelled);
+                }
+                maybe = client_rx.recv() => {
+                    match maybe {
```
lvboudre, line 1162: Make sure the snapshot replay stop if cancel

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1181,19 +1260,14 @@ impl Geyser for GrpcService {
...
-        tokio::spawn(async move {
-            let exit = ping_exit.notified();
-            tokio::pin!(exit);
-
+        let ping_cancellation_token = client_cancelation_token.child_token();
+        self.task_tracker.spawn(async move {
             loop {
                 tokio::select! {
-                    _ = &mut exit => {
+                    _ = ping_cancellation_token.cancelled() => {
```
lvboudre, line 1265: ping task will stop if its child token is canceled.
We are using a child token to allow the client task to cancel all its related "chidlren" subtask spawned by one grpc connection.
This is useful when we want to cancel all of a connection tasks and not the entire plugin tasks.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1227,14 +1301,12 @@ impl Geyser for GrpcService {
         let filter_names = Arc::clone(&self.filter_names);
         let incoming_stream_tx = stream_tx.clone();
         let incoming_client_tx = client_tx;
-        let incoming_exit = Arc::clone(&notify_exit2);
-        tokio::spawn(async move {
-            let exit = incoming_exit.notified();
-            tokio::pin!(exit);
+        let incoming_cancellation_token = client_cancelation_token.child_token();
 
+        self.task_tracker.spawn(async move {
```
lvboudre, line 1303: incoming task will stop if its child token is cancel

Cargo.toml
```diff
@@ -74,6 +74,7 @@ smallvec = "1"
 spl-token-2022 = "8"
 thiserror = "1"
 tokio = "1"
+tokio-util = "0"
```
leafaar, line 77: why 0?

yellowstone-grpc-geyser/src/metrics.rs
```diff
@@ -220,14 +232,16 @@ impl DebugClientStatuses {
 
 #[derive(Debug)]
 pub struct PrometheusService {
+    #[allow(dead_code)]
```
leafaar: why the dead code?

Cargo.toml
lvboudre: to allow any "0.x.y" version. tokio-util does not change much between version.

PR #629, opened 2025-09-28
worksace: remove dependency pinning

lvboudre:


Cargo.toml
```diff
@@ -20,70 +20,70 @@ keywords = ["solana"]
...
+hyper = "^1.4.1"
+hyper-util = "^0.1.7"
+indicatif = "^0.17.9"
+lazy_static = "^1.4.0"
+local-ip-address = "^0.6.1"
+log = "^0.4.17"
+maplit = "^1.0.2"
+prometheus = "^0.13.2"
 prost = "~0.14.0"
 prost-types = "~0.14.0"
```
WilfredAlmeida: Why are `prost` and `prost-types` dependencies `~` if everything else is changed to `^`?

lvboudre: good catch

PR #631, opened 2025-10-08
feat(geyser): optimize account processing latency and add monitoring metrics

leafaar:
## Summary

  Optimizes account processing for lower latency and better observability.

  ## Changes

  - **Remove 10ms timeout**: Account updates arrive rapidly and the timeout was rarely used. Messages
  now send immediately when batches are ready.

  - **Reduce batch size from 31 to 8**: Power-of-2 sizing for better CPU cache alignment and lower
  latency.

  - **Add metrics for visibility**:
    - `account_message_creation_duration_us`: Measures the time it takes to create the message and serialize with protobuf.
    - `account_geyser_loop_update_duration_us`: Time it takes from receiving and creating the message to send on the geyser_loop.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -31,7 +31,7 @@ use {
     tokio::{
         fs,
         sync::{broadcast, mpsc, oneshot, Mutex, RwLock, Semaphore},
-        time::{sleep, Duration, Instant},
+        time::{sleep, Duration},
     },
     tokio_util::{sync::CancellationToken, task::TaskTracker},
     tonic::{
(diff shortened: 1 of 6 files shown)


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -844,6 +844,13 @@ impl GrpcService {
                                     }
                                 }
                             }
+                            if let Message::Account(account) = &message.1 {
+                                let proto_to_systime = SystemTime::try_from(account.created_at).expect("invalid system time");
+                                let elapsed = SystemTime::now()
+                                    .duration_since(proto_to_systime)
+                                    .unwrap_or_default();
```
fanatid, line 851: should we use https://github.com/metrics-rs/quanta from metrics.rs team?

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -844,6 +841,13 @@ impl GrpcService {
...
+                                metrics::observe_account_geyser_loop_update_duration(elapsed);
+                            }
```
leafaar, line 851: I wonder if converting from proto to system adds an overhead here... And how impactful it is?

linuskendall: remove this metric for now! we probably won't be using it.

PR #635, opened 2025-10-25
deps: use spl-token-2022-interface instead of spl-token-2022

fanatid:
In addition removed useless `^` / `~`

I still not sure that it was correct to use next dependencies without `~` because we don't know yet would Anza respect semver for geyser now or not, but release already published so probably we need to stay with default `^` :confused: 
```
agave-geyser-plugin-interface = "3.0.0"
solana-account-decoder = "3.0.0"
solana-storage-proto = "3.0.0"
solana-transaction-context = "3.0.0"
solana-transaction-status = "3.0.0"
```

Everything bumped as patch.

### DIFF
--- a/yellowstone-grpc-client-nodejs/solana-encoding-wasm/Cargo.lock
+++ b/yellowstone-grpc-client-nodejs/solana-encoding-wasm/Cargo.lock
@@ -97,9 +97,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.97"
+version = "1.0.100"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dcfed56ad506cb2c684a14971b8861fdc3baaaae314b9e5f9bb532cbe3ba7a4f"
+checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "arrayref"
@@ -2692,7 +2692,7 @@ dependencies = [
 
 [[package]]
 name = "yellowstone-grpc-proto"
-version = "10.1.0"
+version = "10.1.1"
 dependencies = [
  "anyhow",
  "bincode",


Cargo.toml
```diff
@@ -20,68 +20,68 @@ keywords = ["solana"]
...
+maplit = "1.0.2"
+prometheus = "0.14.0"
+prost = "0.14.0"
+prost-types = "0.14.0"
+prost_011 = { package = "prost", version = "0.11.9" }
+protobuf-src = "1.1.0"
+serde = "1.0.145"
+serde_json = "1.0.86"
+solana-account = "3.0.0"
+solana-account-decoder = "~3.0.4"
```
lvboudre: remove the "~" ?

Cargo.toml
```diff
@@ -20,68 +20,68 @@ keywords = ["solana"]
...
+tokio-stream = "0.1.17"
+tokio-util = "0.7.0"
+tonic = "0.14.0"
+tonic-build = "0.14.0"
+tonic-prost = "0.14.0"
+tonic-prost-build = "0.14.0"
+tonic-health = "0.14.0"
+vergen = "9.0.6"
 yellowstone-grpc-client = { path = "yellowstone-grpc-client", version = "10.0.0" }
 yellowstone-grpc-proto = { path = "yellowstone-grpc-proto", version = "10.0.0", default-features = false }
```
lvboudre: You will need to put 10.1.1 I guess

lvboudre (changes requested):

Cargo.toml
fanatid: fixed, thanks

Cargo.toml
fanatid: No, `10.1.1` is not required here, everything should work if you only change spl-token-2022 -> spl-token-2022-interface (but if you require new feature from 10.1.0 this can requires bump)

fanatid: bumped to 10.1.0 because new features introduced (again 10.1.0 with token2022 in 10.1.1 are safe)

lvboudre (approved):

PR #640, opened 2025-11-10
fix: Reset metrics on plugin unload

### LINKED ISSUE #639: unloading and loading plugin doesn't reset metrics
The metrics for connection counts do not seem to be getting reset when the plugin is unloaded and loaded

For instance say I create 6 subscriptions.

I check the metrics, and I'll see

```
subscriptions_total{endpoint="",subscription="grpc_total"} 6
```

Then I unload the plugin.  Connections are terminated.

I load the plugin again, and create 1 subscription:

I check the metrics, and I'll see
```
subscriptions_total{endpoint="",subscription="grpc_total"} 7
```

I would expect to see grpc_total at 6... not 7.

joe-hellomoon:
Fixes #639

Metrics accumulate across plugin reload cycles because they're stored in static variables that persist for the process lifetime.

Added reset_metrics() function that clears gauges and counters when the plugin is unloaded.

### DIFF
--- a/yellowstone-grpc-geyser/src/metrics.rs
+++ b/yellowstone-grpc-geyser/src/metrics.rs
@@ -477,3 +477,28 @@ pub fn set_subscriber_queue_size<S: AsRef<str>>(subscriber_id: S, size: u64) {
         .with_label_values(&[subscriber_id.as_ref()])
         .set(size as i64);
 }
+
+/// Reset all metrics on plugin unload to prevent metric accumulation across plugin lifecycle
+pub fn reset_metrics() {
+    // Reset gauge metrics to 0
+    CONNECTIONS_TOTAL.set(0);
+    MESSAGE_QUEUE_SIZE.set(0);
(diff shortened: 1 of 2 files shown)


leafaar (approved):
Thanks for the PR 😁 Looks good to me!

leafaar:
@joe-hellomoon Can you fix the conflicts please? Also, I would like to request @lvboudre review here.

leafaar:
@joe-hellomoon The tests didn't pass, apparently there's no method `reset` on the `GEYSER_ACCOUNT_UPDATE_RECEIVED`, could you check that please?

joe-hellomoon:
> @joe-hellomoon The tests didn't pass, apparently there's no method `reset` on the `GEYSER_ACCOUNT_UPDATE_RECEIVED`, could you check that please?

yeah apparently you can't reset histograms after all.. :-\ 

ah well, reset removed!

PR #641, opened 2025-11-13
group solana deps

grooviegermanikus:
Motivation: When updating Solana dependencies the Monorepo version sometimes diverges from solana-sdk and maybe others which makes it harder to choose the right version.

Group solana dependencies in _Cargo.toml_ to make clear which dependencies are pulled from Anza monorepo vs solana-sdk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5064,7 +5064,6 @@ dependencies = [
  "solana-clock",
  "solana-logger",
  "solana-pubkey",
- "solana-transaction-status",
  "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,6 @@ publish = false
 
 [workspace.dependencies]
 affinity = "0.1.2"
-agave-geyser-plugin-interface = "3.0.0"
 anyhow = "1.0.98"
 backoff = "0.4.0"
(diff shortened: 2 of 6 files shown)


leafaar:
I see you converted this into a draft, feel free to tag me when you're ready for a review!

leafaar:
@grooviegermanikus I think you did a `cargo update` and some dependencies got updated. You can try and roll back to see if it works. Or you can also manually downgrade then (i.e. `cargo update crate-name --precise version`

grooviegermanikus:
unfortunately the project cannot be built on MacOS:

```
error: non-item macro in item position: ::core::panic
  --> /Users/me/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/affinity-0.1.2/src/lib.rs:43:9
   |
43 |         unimplemented!("This crate does not support your OS yet !");

```

lmk if you want/can fix that. as a workaround I switched to a different cpu-lib. or we can make affinity a no-no on macOS

leafaar:
It was just a thing to make it work while I was testing out things, but I was not so sure about the code... I was thinking about this PR, maybe it's worth merging it because most devs use Mac... I will check that code again and fix some issues, but I will need till Monday so other dev can test/review.

leafaar:
I will need you to sign the commits so we can merge this one @grooviegermanikus :grin:

PR #642, opened 2025-11-14
fixes lifetime capture issue for Rust edition 2024

grooviegermanikus:
this is the report from rust compiler (1.86):
> note: this call may capture more lifetimes than intended, because Rust 2024 has adjusted the `impl Trait` lifetime capture rules
> help: if you can modify this crate, add a precise capturing bound to avoid overcapturing: `+ use<F>`

this is the usage:
```rust
let fut_subscribe = timeout(
  subscribe_timeout,
  client.subscribe_with_request(Some(subscribe_filter_on_connect)),
);
```

disclaimer: I've just read 2 minutes about that issue so maybe what I'm proposing is totally wrong


leafaar:
@grooviegermanikus Can you sign the commit so we can merge this one?

PR #644, opened 2025-11-20
fix: variable name typos in `grpc.rs`

darth-raijin:
This PR fixes two variable name typos in `grpc.rs`. These variables are locally scoped, so the change is not breaking any API.


darth-raijin:
@leafar just checking in, is this good to merge?

PR #645, opened 2025-11-27
made GeyserGrpcClient implements Clone trait

### LINKED ISSUE #584: Opaque `Interceptor`  in `GrpcGeyserClient` prevents cloning of client

lvboudre:
Fixes #584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,18 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Breaking
 
+## 2025-11-27
+
+- yellowstone-grpc-client-simple-10.1.1
+- yellowstone-grpc-client-10.1.1
+- yellowstone-grpc-geyser-10.2.0
+- yellowstone-grpc-proto-10.1.1
+
+### Features
+
+- Made `GeyserGrpcClient` implements `Clone`
+
+
 ## 2025-10-25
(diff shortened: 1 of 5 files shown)


leafaar (approved):
lgtm

PR #650, opened 2025-12-08
Add support for preprocessed messages 

pmantica11:
# Overview 

In this PR, I add support for preprocessed transactions to the gRPC Yellowstone server, making it a unified API for all Solana data needs. To support preprocessed transactions, we rely on using a Jito shredstream proxy to consume decoded transactions. Alternatively, we could add the filtering capabilities into the shredstream proxy itself. However, it's neater to expose this in the Yellowstone gRPC to keep YS as the fan-out layer. 

Next steps:
1. Add documentation and client SDK in upcoming PRs. 

# Testing

Added unit tests for filtering and tested locally that everything works e2e.


yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -18,6 +19,41 @@ service Geyser {
...
+// Subscriptions to Preprocessed transactions
+message SubscribePreprocessedRequest {
+  map<string, SubscribePreprocessedRequestFilterTransactions> transactions = 1;
+  optional SubscribeRequestPing ping = 2;
+}
+
+message SubscribePreprocessedRequestFilterTransactions {
+  optional bool vote = 1;
+  optional string signature = 2;
+  repeated string account_include = 3;
```
linuskendall, line 31: These might be better renamed as `static_account_include`, `static_account_exclude`, `static_account_required` as they are not the same filters as the account exclude/required/include for deshredded transactions as they would for processed transactions.

PR #653, opened 2026-01-12
Pre-encode transaction content to reduce serialization overhead

Ozodimgba:
Closes #STR-258
## Problem

Each connected client triggers independent encoding of the same transaction data. Prost's two-pass encoding (length calculation + byte writing) means N clients cause 2N full traversals of the transaction struct. Profiling on mainnet showed 12-15% CPU in encoding-related functions (`tx_encoded_len`, `put_slice`, `encode_varint`).

## Solution

Pre-encode transaction content once when `MessageTransactionInfo` is created. Store the bytes alongside the struct. When encoding for clients, copy the pre-computed bytes instead of re-encoding.

## Changes

- Add `pre_encoded: Option<Bytes>` field to `MessageTransactionInfo`
- Call `pre_encode()` in `from_geyser()` and `from_update_oneof()`
- Modify `tx_encode_raw()` to use pre-encoded bytes when available
- Modify `tx_encoded_len()` to return cached length when available
- Fallback to manual encoding when `pre_encoded` is `None` (tests, edge cases)

## Why this works

Transaction filters are gatekeepers, not transformers. A transaction either passes the filter (sent whole) or doesn't (dropped). The encoded bytes are identical for every client who receives the transaction. Only the small `filters` wrapper field differs per client.

## Testing

- `test_pre_encoded_matches_manual_encoding`: verifies pre-encoded bytes match
  manual encoding for 308 real transactions from fixtures
- All existing encode/decode round-trip tests pass

## Expected impact

| Metric | Before | After |
|--------|--------|-------|
| Encoding passes per tx | N × 2 | 1 + N copies |
| `tx_encoded_len` calls | N per tx | 1 per tx |

Actual CPU reduction to be measured via profiling.

### DIFF
--- a/yellowstone-grpc-proto/src/plugin/filter/message.rs
+++ b/yellowstone-grpc-proto/src/plugin/filter/message.rs
@@ -288,6 +288,7 @@ impl FilteredUpdate {
                         },
                         index: msg.index as usize,
                         account_keys: HashSet::new(),
+                        pre_encoded: None,
                     }),
                     slot: msg.slot,
                 })
@@ -661,6 +662,15 @@ impl prost::Message for FilteredUpdateTransaction {
 
 impl FilteredUpdateTransaction {
     fn tx_encode_raw(tag: u32, tx: &MessageTransactionInfo, buf: &mut impl BufMut) {
+        // try to use pre-encoded bytes (fast path)
+        if let Some(pre_encoded) = tx.get_pre_encoded() {
+            encode_key(tag, WireType::LengthDelimited, buf);
+            encode_varint(pre_encoded.len() as u64, buf);
+            buf.put_slice(pre_encoded);
+            return;
+        }
+
+        // fallback: encode from scratch
         encode_key(tag, WireType::LengthDelimited, buf);
         encode_varint(Self::tx_encoded_len(tx) as u64, buf);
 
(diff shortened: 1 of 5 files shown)


yellowstone-grpc-proto/src/plugin/filter/message.rs
```diff
@@ -661,6 +662,15 @@ impl prost::Message for FilteredUpdateTransaction {
 
 impl FilteredUpdateTransaction {
     fn tx_encode_raw(tag: u32, tx: &MessageTransactionInfo, buf: &mut impl BufMut) {
+        // try to use pre-encoded bytes (fast path)
+        if let Some(pre_encoded) = tx.get_pre_encoded() {
+            encode_key(tag, WireType::LengthDelimited, buf);
+            encode_varint(pre_encoded.len() as u64, buf);
+            buf.put_slice(pre_encoded);
+            return;
+        }
```
lvboudre, line 671: I like that you made this optional and fallback to default routine.

yellowstone-grpc-proto/src/plugin/message.rs
```diff
@@ -353,6 +362,71 @@ impl MessageTransactionInfo {
         self.account_keys = account_keys;
         Ok(())
     }
+
+    #[inline]
+    pub fn get_pre_encoded(&self) -> Option<&Bytes> {
+        self.pre_encoded.as_ref()
+    }
+
+    pub fn pre_encode(&mut self) {
```
lvboudre: where in the plugin lifecycle will we call this function?

lvboudre: Same as pre-encoded accout, design-wise its not good to add encoding responsability to a data object.
An object should not know how to encode itself -- what if the encoding change? what if the framework change? what if we want to support other kind of encoding.
It's better to move it as a utility function otherwise the `MessageTransactionInfo`

PR #654, opened 2026-01-14
fix 3.1 upgrade

fanatid:


Cargo.toml
```diff
@@ -82,8 +65,33 @@ tonic-prost = "0.14.0"
...
-yellowstone-grpc-client = { path = "yellowstone-grpc-client", version = "10.1.0" }
-yellowstone-grpc-proto = { path = "yellowstone-grpc-proto", version = "10.1.0", default-features = false }
+
+# Agave Monorepo
+agave-geyser-plugin-interface = "~3.1.0"
+solana-account-decoder = "~3.1.0"
+solana-logger = "3.0.0"
+solana-storage-proto = { version = "~3.1.0", features = ["agave-unstable-api"] }
+solana-transaction-context = "~3.1.0"
+solana-transaction-status = "~3.1.0"
```
lvboudre, line 75: Is it necessary to pin those?

Cargo.toml
```diff
@@ -1,13 +1,13 @@
 [workspace]
 resolver = "2"
 members = [
-    "examples/rust", # 10.1.1
-    "yellowstone-grpc-client", # 10.1.1
-    "yellowstone-grpc-geyser", # 10.1.1
-    "yellowstone-grpc-proto", # 10.1.1
+    "examples/rust", # 11.0.0
+    "yellowstone-grpc-client", # 11.0.0
```
lvboudre, line 5: Why major bump?

Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
+    "yellowstone-grpc-geyser", # 11.0.0
+    "yellowstone-grpc-proto", # 11.0.0
```
lvboudre, line 7: same: why major bump?

Cargo.toml
leafaar: they are not pinned

fanatid: anza/agave still consider minor update as huge step (that sometimes have backward compatibility?), I'd like to see that plugin follows agave updates
`~` don't pin it completely (compare to `=`), only restrict versions, i.e. same as `>=3.1.0,<3.2.0`

fanatid: I didn't put modifier to `solana-logger` (and `^` is used by default) because `solana-logger` don't have dependencies to any solana crate

lvboudre: @leafaar yes putting this "~" pins the minor version.

Cargo.toml
```diff
@@ -82,8 +65,33 @@ tonic-prost = "0.14.0"
...
 tonic-health = "0.14.0"
 vergen = "9.0.6"
-yellowstone-grpc-client = { path = "yellowstone-grpc-client", version = "10.1.0" }
-yellowstone-grpc-proto = { path = "yellowstone-grpc-proto", version = "10.1.0", default-features = false }
+
+# Agave Monorepo
+agave-geyser-plugin-interface = "~3.1.0"
+solana-account-decoder = "~3.1.0"
+solana-logger = "3.0.0"
+solana-storage-proto = { version = "~3.1.0", features = ["agave-unstable-api"] }
```
lvboudre, line 73: can we remove `features = ["agave-unstable-api"]` and only add it in individual crate that needs it?

lvboudre (approved):

PR #655, opened 2026-01-16
Add deshred transaction subscription support

leafaar:
Introduces support for subscribing to deshred transactions, which are transactions received when entries are formed from shreds, before any execution occurs.

Key characteristics of deshred transactions:
- No TransactionStatusMeta available (no execution results yet)
- Address lookup tables have NOT been resolved
- Only static account keys from the transaction message are available

Proto changes:
- SubscribeRequestFilterDeshredTransactions with vote filter and static_account_include/exclude fields
- SubscribeUpdateDeshredTransaction and SubscribeUpdateDeshredTransactionInfo

Filter implementation:
- FilterDeshredTransactions with vote filtering
- Static account key filtering (include/exclude)
- FilterLimitsDeshredTransactions for rate limiting configuration

Plugin changes:
- notify_deshred_transaction hook
- deshred_transaction_notifications_enabled flag

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -67,7 +67,7 @@ tonic-health = "0.14.0"
 vergen = "9.0.6"
 
 # Agave Monorepo
-agave-geyser-plugin-interface = "3.0.0"
+agave-geyser-plugin-interface = { git = "https://github.com/rpcpool/solana-public", branch = "v3.0-triton-public"}
 solana-account-decoder = "3.0.0"
 solana-logger = "3.0.0"
 solana-storage-proto = "3.0.0"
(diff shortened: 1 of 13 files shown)


yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -42,6 +42,7 @@ message SubscribeRequest {
   map<string, SubscribeRequestFilterBlocks> blocks = 4;
   map<string, SubscribeRequestFilterBlocksMeta> blocks_meta = 5;
   map<string, SubscribeRequestFilterEntry> entry = 8;
+  map<string, SubscribeRequestFilterDeshredTransactions> deshred_transactions = 12;
```
lvboudre: We wont be introducing it in `subcribe` and not inside `SubscribeRequest`.
Lets build another method called `subscribeDeshred` and `SubscribeDeshredRequest` type for this optional feature.

Cargo.toml
```diff
@@ -67,7 +67,7 @@ tonic-health = "0.14.0"
 vergen = "9.0.6"
 
 # Agave Monorepo
-agave-geyser-plugin-interface = "3.0.0"
+agave-geyser-plugin-interface = { git = "https://github.com/rpcpool/solana-public", branch = "deshred-transaction"}
```
lvboudre: We should create a dedicated branch so the official plubic yellowstone-grpc does not force anyone to use our forks for the first release.

PR #657, opened 2026-01-19
Pre-encode account content to reduce serialization overhead

Ozodimgba:
Closes #STR-287
<html><head></head><body><h2>Pre-encode account content to reduce serialization overhead</h2>
<h3>Problem</h3>
<p>Each connected client triggers independent encoding of the same account data. Prost's two-pass encoding (length calculation + byte writing) means N clients cause 2N full traversals of the account struct. Accounts are 3x more frequent than transactions (~6k accounts vs ~2k transactions per slot), making this a significant source of encoding overhead.</p>
<h3>Solution</h3>
<p>Pre-encode account content once when <code>MessageAccountInfo</code> is created. Store the bytes alongside the struct. When encoding for clients <strong>without data slicing</strong>, copy the pre-computed bytes instead of re-encoding.</p>
<h3>Changes</h3>
<ul>
<li>Add <code>pre_encoded: Option&lt;Bytes&gt;</code> field to <code>MessageAccountInfo</code></li>
<li>Call <code>pre_encode()</code> in <code>from_geyser()</code> and <code>from_update_oneof()</code></li>
<li>Modify <code>account_encode_raw()</code> to use pre-encoded bytes when <code>data_slice.is_empty()</code></li>
<li>Modify <code>account_encoded_len()</code> to return cached length when applicable</li>
<li>Fallback to manual encoding when:
<ul>
<li><code>pre_encoded</code> is <code>None</code> (tests, edge cases)</li>
<li>Client requests data slicing (slice-aware encoding required)</li>
</ul>
</li>
</ul>
<h3>Why this works</h3>
<p>For clients requesting full account data (no slicing), the encoded bytes are identical. We can share pre-computed bytes across these clients. Clients with data slicing need custom byte ranges, so they fall back to the existing slice-aware encoding path.</p>
<pre><code class="rust">// In account_encode_raw:
if data_slice.as_ref().is_empty() {
    if let Some(pre_encoded) = account.get_pre_encoded() {
        // Fast path: copy pre-encoded bytes
        buf.put_slice(pre_encoded);
        return;
    }
}
// Slow path: slice-aware encoding
</code></pre>
<h3>Testing</h3>
<ul>
<li><code>test_account_pre_encoded_matches_manual_encoding</code>: verifies pre-encoded bytes match manual encoding across various account configurations (different lamports, data sizes, executability, signatures)</li>
<li>All existing encode/decode round-trip tests pass</li>
</ul>
<h3>Expected impact</h3>
<p>For clients without data slicing:</p>

Metric | Before | After
-- | -- | --
Encoding passes per account | N × 2 | 1 + N copies
account_encoded_len calls | N per account | 1 per account

<p>Clients with data slicing: no change (existing behavior preserved).</p>

### DIFF
--- a/yellowstone-grpc-proto/src/plugin/message.rs
+++ b/yellowstone-grpc-proto/src/plugin/message.rs
@@ -7,7 +7,7 @@ use {
             SubscribeUpdateBlock, SubscribeUpdateBlockMeta, SubscribeUpdateEntry,
             SubscribeUpdateSlot, SubscribeUpdateTransaction, SubscribeUpdateTransactionInfo,
         },
-        plugin::filter::encoder::TransactionEncoder,
+        plugin::filter::encoder::{AccountEncoder, TransactionEncoder},
         solana::storage::confirmed_block,
     },
     agave_geyser_plugin_interface::geyser_plugin_interface::{
@@ -195,13 +195,14 @@ pub struct MessageAccountInfo {
     pub data: Bytes,
     pub write_version: u64,
     pub txn_signature: Option<Signature>,
+    pub pre_encoded: Option<Bytes>,
 }
 
 impl MessageAccountInfo {
     pub fn from_geyser(info: &ReplicaAccountInfoV3<'_>) -> Self {
         let shared = info.data.to_vec();
         let data = Bytes::from(shared);
-        Self {
+        let mut account = Self {
             pubkey: Pubkey::try_from(info.pubkey).expect("valid Pubkey"),
             lamports: info.lamports,
             owner: Pubkey::try_from(info.owner).expect("valid Pubkey"),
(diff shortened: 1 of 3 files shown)


perf-full.txt
lvboudre, line 1: Can you remove this file please, it should not be part of the public repository.

yellowstone-grpc-proto/src/plugin/message.rs
```diff
@@ -230,7 +229,75 @@ impl MessageAccountInfo {
...
+        }
+        if self.write_version != 0u64 {
+            ::prost::encoding::uint64::encode(7u32, &self.write_version, &mut buf);
+        }
+        if let Some(value) = &self.txn_signature {
+            prost_bytes_encode_raw(8u32, value.as_ref(), &mut buf);
+        }
+
+        self.pre_encoded = Some(Bytes::from(buf));
+    }
```
lvboudre: Design-wise an object should not know how to encode itself,  we should have dedicated object whose responsibility is to encode. If the encoding change or the framework for encoding it can introduce bug in what-seems unrelated objects.

Can we move this function as a utility function.

perf-full.txt
Ozodimgba: yeah will remove it

yellowstone-grpc-proto/src/plugin/message.rs
Ozodimgba: Moving pre_encode() to a utility function on FilteredUpdateTransaction, alongside the existing tx_encode_raw() and tx_encoded_len().

PR #681, opened 2026-02-11
feat: increase default HTTP/2 window sizes to 32MB

pmantica11:
## Summary

Increase the default HTTP/2 stream and connection window sizes to 32MB across all clients to improve throughput on high-latency connections, and document recommended OS-level TCP tuning in the README.

### Motivation

Benchmarking over an intercontinental connection (Pittsburgh, PA → Frankfurt, Germany) showed that HTTP/2 flow control window sizes are a major throughput bottleneck. Increasing the window size to 32MB provides optimal throughput without excessive memory overhead.

#### Throughput benchmark: HTTP/2 window size vs TCP buffer size

Measured using LaserStream replay (subscribing to all accounts, slots, transactions, transaction statuses, entries, and blocks meta from `current_slot - 300`) over a Pittsburgh-to-Frankfurt intercontinental connection, 5 seconds per configuration:

| TCP Buffer ↓ \ Window → | hyper default | 64KB | 4MB | 8MB | 16MB | 32MB | 64MB |
|---|---|---|---|---|---|---|---|
| **4MB** | 14.67 MB/s | 0.33 MB/s | 28.25 MB/s | 26.77 MB/s | 28.19 MB/s | 26.73 MB/s | 28.29 MB/s |
| **8MB** | 14.48 MB/s | 0.31 MB/s | 28.72 MB/s | 52.44 MB/s | 55.13 MB/s | 52.42 MB/s | 52.59 MB/s |
| **16MB** | 14.68 MB/s | 0.55 MB/s | 28.63 MB/s | 55.90 MB/s | 102.48 MB/s | 94.11 MB/s | 101.99 MB/s |
| **32MB** | 14.66 MB/s | 0.55 MB/s | 28.68 MB/s | 55.97 MB/s | 99.51 MB/s | **118.38 MB/s** | **124.47 MB/s** |
| **64MB** | 14.57 MB/s | 0.54 MB/s | 28.79 MB/s | 54.81 MB/s | 100.43 MB/s | 100.74 MB/s | 102.22 MB/s |

Key findings:
- **hyper default** caps at ~14.6 MB/s regardless of TCP buffer size. Hyper's default starts at the HTTP/2 spec value (65535 bytes) but uses adaptive flow control that grows the window dynamically — explaining why it performs ~30x better than an explicit 64KB setting
- **Explicit 64KB** (~0.5 MB/s) hard-locks the window at 65535 bytes with no adaptation, confirming flow control is the bottleneck
- **Sweet spot is 32MB TCP + 32-64MB window** at ~118-124 MB/s
- 64MB TCP buffer performs *worse* than 32MB (likely buffer bloat / memory pressure)
- Throughput roughly doubles at each window size step: 4MB→8MB→16MB, then plateaus
- 32MB window provides 8x improvement over hyper's default with minimal memory overhead

### Changes

**Rust client library** (`yellowstone-grpc-client`):
- `initial_stream_window_size`: unset (hyper default) → 32MB
- `initial_connection_window_size`: unset (hyper default) → 32MB
- Defaults applied in `from_shared()`/`from_static()`, not in `new()`, so clients that construct the builder with a pre-configured `Endpoint` are not overridden

**Node.js client** (`yellowstone-grpc-client-nodejs`):
- `grpc_initial_connection_window_size`: 8MB → 32MB
- `grpc_initial_stream_window_size`: 4MB → 32MB

**Rust example client**:
- `--initial-stream-window-size` and `--initial-connection-window-size` are now optional
- If not specified, uses library default (32MB). Previously always overrode with 64MB.

**Go example client**:
- `grpc.WithInitialWindowSize(32MB)` and `grpc.WithInitialConnWindowSize(32MB)`

**Python example client**:
- Added `grpc.initial_window_size` and `grpc.initial_connection_window_size` (32MB)
- Previously unset, using gRPC-Python's default

**README**:
- Updated "System Tuning" section with recommended `sysctl` settings for 32MB TCP receive buffers

### Trade-offs

32MB window sizes allow the sender to push ~32MB of data before waiting for flow control ACKs, which is critical on high-latency links. The memory overhead (~32MB worst case per stream) is negligible for most use cases. The benchmarks show diminishing returns above 32MB, making it the optimal default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
+++ b/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
@@ -56,8 +56,8 @@ impl Default for JsChannelOptions {
   /// JsChannelOptions {
   ///   grpc_connect_timeout: Duration::from_secs(10),
   ///   grpc_timeout: Some(Duration::from_secs(30)),
-  ///   grpc_initial_connection_window_size: Some(64 * 1024 * 1024), // 64MB
-  ///   grpc_initial_stream_window_size: Some(64 * 1024 * 1024),     // 64MB
+  ///   grpc_initial_connection_window_size: Some(32 * 1024 * 1024), // 32MB
+  ///   grpc_initial_stream_window_size: Some(32 * 1024 * 1024),     // 32MB
   ///   grpc_max_decoding_message_size: Some(1 * 1024 * 1024 * 1024), // 1GB
   ///   grpc_max_encoding_message_size: Some(32 * 1024 * 1024),     // 32MB
   ///   grpc_http2_adaptive_window: Some(true),
@@ -75,8 +75,8 @@ impl Default for JsChannelOptions {
     Self {
       grpc_connect_timeout: Some(10_000),
       grpc_timeout: Some(30_000),
-      grpc_initial_connection_window_size: Some(64 * 1024 * 1024), // 64MB
-      grpc_initial_stream_window_size: Some(64 * 1024 * 1024),     // 64MB
+      grpc_initial_connection_window_size: Some(32 * 1024 * 1024), // 32MB
+      grpc_initial_stream_window_size: Some(32 * 1024 * 1024),     // 32MB
       grpc_max_decoding_message_size: Some(1024 * 1024 * 1024),   // 1GB
       grpc_max_encoding_message_size: Some(32 * 1024 * 1024),     // 32MB
       grpc_http2_adaptive_window: Some(true),
--- a/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
+++ b/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
@@ -56,8 +56,8 @@ impl Default for JsChannelOptions {
   /// JsChannelOptions {
   ///   grpc_connect_timeout: Duration::from_secs(10),
   ///   grpc_timeout: Some(Duration::from_secs(30)),
-  ///   grpc_initial_connection_window_size: Some(8 * 1024 * 1024), // 8MB
-  ///   grpc_initial_stream_window_size: Some(4 * 1024 * 1024),     // 4MB
+  ///   grpc_initial_connection_window_size: Some(64 * 1024 * 1024), // 64MB
+  ///   grpc_initial_stream_window_size: Some(64 * 1024 * 1024),     // 64MB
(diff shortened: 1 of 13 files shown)


yellowstone-grpc-client-nodejs/napi/src/bindings.rs
```diff
@@ -75,8 +75,8 @@ impl Default for JsChannelOptions {
     Self {
       grpc_connect_timeout: Some(10_000),
       grpc_timeout: Some(30_000),
-      grpc_initial_connection_window_size: Some(8 * 1024 * 1024), // 8MB
-      grpc_initial_stream_window_size: Some(4 * 1024 * 1024),     // 4MB
+      grpc_initial_connection_window_size: Some(32 * 1024 * 1024), // 32MB
+      grpc_initial_stream_window_size: Some(32 * 1024 * 1024),     // 32MB
```
leafaar, line 79:
```suggestion
      grpc_initial_connection_window_size: Some(32 * 1024 * 1024), // 32MiB
      grpc_initial_stream_window_size: Some(32 * 1024 * 1024),     // 32MiB
```
Could we rename other occurrences to MiB for consistency?
We represent sizes in octets (base-2), so MiB would be technically correct instead of MB.

lvboudre:
I appreciate the PR, but we still need tcpdump and Wireshark traces to confirm whether a 32 MiB window size is actually feasible.

I’m not comfortable setting this as the default, since it effectively optimizes for the worst-case scenario. I’d prefer this to be optional. Most users are not connecting across continents, and in most cases it’s better to let the adaptive window handle sizing.

For users with ~100 ms latency, enabling Zstd compression should significantly reduce window size requirements, as it lowers bandwidth usage by roughly a factor of 2.2. Optimizing by default for 100 ms latency in low-latency streaming is generally an anti-pattern, and we shouldn’t design defaults around it.

We’ve already run cross-continent US–EU experiments with latency above 100 ms, and with Zstd enabled, the issue is resolved.

Finally, setting a very large window size can make connections more fragile in the presence of TCP packet loss. When packet loss occurs with a large window and adaptive sizing enabled, the system may aggressively reduce the window for an extended period. This can cause the remote client to fall behind in the Geyser plugin and eventually terminate the connection.

lvboudre:
Closed because I don't hear from you since the recent conversation we had in slack + I can't reproduce you supposedly zstd limitation.

PR #717, opened 2026-03-23
STR-404: Yellowstone-grpc-client stream auto-reconnect

Ozodimgba:
Closes #STR-404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,12 @@ The minor version will be incremented upon a breaking change and the patch versi
 ## 2026-04-01
 
 - yellowstone-grpc-client-13.1.0
-- yellowstone-grpc-client-simple-13.1.0
+- yellowstone-grpc-client-simple-12.3.0
 
 ### Features
 
 - client: auto-reconnect stream for yellowstone-grpc-client ([#717](https://github.com/rpcpool/yellowstone-grpc/pull/717))
+- geyser: set replay_stored_slots default to 150, added startup warning for low values ([#717](https://github.com/rpcpool/yellowstone-grpc/pull/717))
 
 ## 2026-04-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 **Note:** Version 0 of Semantic Versioning is handled differently from version 1 and above.
 The minor version will be incremented upon a breaking change and the patch version will be incremented for features.
+## 2026-04-01
+
+- yellowstone-grpc-client-13.1.0
+- yellowstone-grpc-client-simple-13.1.0
+
+### Features
+
+- client: auto-reconnect stream for yellowstone-grpc-client ([#717](https://github.com/rpcpool/yellowstone-grpc/pull/717))
 
 ## 2026-04-10
 
--- a/examples/rust/src/bin/client_with_reconnect.rs
+++ b/examples/rust/src/bin/client_with_reconnect.rs
@@ -1,0 +1,111 @@
+use {
+    clap::Parser,
+    futures::stream::StreamExt,
+    log::info,
+    std::collections::HashMap,
+    yellowstone_grpc_client::{GeyserGrpcClient, ReconnectConfig},
+    yellowstone_grpc_proto::prelude::{
+        subscribe_update::UpdateOneof, CommitmentLevel, SubscribeRequest,
+        SubscribeRequestFilterAccounts, SubscribeRequestFilterSlots,
+        SubscribeRequestFilterTransactions,
+    },
+};
+
+#[derive(Debug, Clone, Parser)]
+#[clap(author, version, about = "Yellowstone gRPC client with auto-reconnect")]
+struct Args {
+    #[clap(short, long, default_value_t = String::from("http://127.0.0.1:10000"))]
+    endpoint: String,
+
+    #[clap(long)]
+    x_token: Option<String>,
+
+    #[clap(long)]
+    slots: bool,
+
+    #[clap(long)]
+    accounts: bool,
+
+    #[clap(long)]
+    transactions: bool,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    env_logger::init();
+    let args = Args::parse();
+
+    let mut client = GeyserGrpcClient::build_from_shared(args.endpoint)?
+        .x_token(args.x_token)?
+        .set_reconnect_config(ReconnectConfig::default())
+        .connect()
+        .await?;
+
+    let request = SubscribeRequest {
+        slots: if args.slots {
+            let mut m = HashMap::new();
+            m.insert(
+                "".to_owned(),
+                SubscribeRequestFilterSlots {
+                    filter_by_commitment: Some(true),
+                    interslot_updates: None,
+                },
+            );
+            m
+        } else {
+            HashMap::new()
+        },
+        accounts: if args.accounts {
+            let mut m = HashMap::new();
+            m.insert("".to_owned(), SubscribeRequestFilterAccounts::default());
+            m
+        } else {
+            HashMap::new()
+        },
+        transactions: if args.transactions {
+            let mut m = HashMap::new();
+            m.insert("".to_owned(), SubscribeRequestFilterTransactions::default());
+            m
+        } else {
+            HashMap::new()
+        },
+        commitment: Some(CommitmentLevel::Processed as i32),
+        ..Default::default()
+    };
+
+    info!("connecting with auto-reconnect enabled");
+    let mut stream = client.subscribe_once(request).await?;
+    let mut count = 0u64;
+
+    while let Some(msg) = stream.next().await {
+        match msg {
+            Ok(update) => {
+                count += 1;
+                match update.update_oneof.as_ref() {
+                    Some(UpdateOneof::Slot(slot)) => {
+                        if count % 10 == 0 {
+                            info!("slot={} count={count}", slot.slot);
+                        }
+                    }
+                    Some(UpdateOneof::Account(acc)) => {
+                        if count % 100 == 0 {
+                            info!("account update slot={} count={count}", acc.slot);
+                        }
+                    }
+                    Some(UpdateOneof::Transaction(tx)) => {
+                        if count % 100 == 0 {
+                            info!("transaction slot={} count={count}", tx.slot);
+                        }
+                    }
+                    _ => {}
+                }
+            }
+            Err(status) => {
+                log::error!("fatal: {status}");
+                break;
+            }
+        }
+    }
+
+    Ok(())
+}
--- a/yellowstone-grpc-client/src/lib.rs
+++ b/yellowstone-grpc-client/src/lib.rs
@@ -1,13 +1,23 @@
-pub use tonic::{service::Interceptor, transport::ClientTlsConfig};
+mod dedup;
+mod reconnect;
+
 use {
+    crate::{
+        dedup::DEFAULT_SLOT_RETENTION,
+        reconnect::{TonicGeyserClientOptions, AUTORECONNECT_FILTER_KEY},
+    },
+    arc_swap::ArcSwap,
     bytes::Bytes,
     futures::{
         channel::mpsc,
         sink::{Sink, SinkExt},
         stream::Stream,
-        StreamExt,
     },
-    std::{path::PathBuf, time::Duration},
+    std::{
+        path::PathBuf,
+        sync::{Arc, Mutex},
+        time::Duration,
+    },
     tokio::net::UnixStream,
     tonic::{
         codec::{CompressionEncoding, Streaming},
@@ -29,6 +39,16 @@ use {
         SubscribeRequest, SubscribeUpdate, SubscribeUpdateDeshred,
     },
 };
+pub use {
+    crate::{
+        dedup::{DedupState, DedupStream},
+        reconnect::{AutoReconnect, Backoff, GrpcConnector, TonicGrpcConnector},
+    },
+    tonic::{service::Interceptor, transport::ClientTlsConfig},
+};
+
+#[cfg(feature = "test-tools")]
+pub mod test_tools;
 
 #[derive(Debug, Clone)]
 pub struct InterceptorXToken {
@@ -54,9 +74,46 @@ impl Interceptor for InterceptorXToken {
 pub enum GeyserGrpcClientError {
     #[error("gRPC status: {0}")]
     TonicStatus(#[from] Status),
+    #[error("gRPC transport error: {0}")]
+    TransportError(#[from] tonic::transport::Error),
 }
 
 pub type GeyserGrpcClientResult<T> = Result<T, GeyserGrpcClientError>;
+
+#[derive(Clone, Debug)]
+/// Configuration for automatic subscribe reconnect behavior.
+pub struct ReconnectConfig {
+    pub backoff: Backoff,
+    pub slot_retention: usize,
+}
+
+impl Default for ReconnectConfig {
+    fn default() -> Self {
+        Self {
+            backoff: Backoff::default(),
+            slot_retention: DEFAULT_SLOT_RETENTION,
+        }
+    }
+}
+
+impl ReconnectConfig {
+    pub const fn no_reconnect() -> Self {
+        Self {
+            backoff: Backoff::new(Duration::from_millis(0), 1.0, 0),
+            slot_retention: 0,
+        }
+    }
+
+    pub const fn with_backoff(mut self, backoff: Backoff) -> Self {
+        self.backoff = backoff;
+        self
+    }
+
+    pub const fn with_slot_retention(mut self, slot_retention: usize) -> Self {
+        self.slot_retention = slot_retention;
+        self
+    }
+}
 
 ///
 /// See [`GeyserGrpcBuilder`] for constructing a client with custom options.
@@ -65,6 +122,10 @@ pub type GeyserGrpcClientResult<T> = Result<T, GeyserGrpcClientError>;
 pub struct GeyserGrpcClient {
     pub health: HealthClient<InterceptedService<Channel, InterceptorXToken>>,
     pub geyser: GeyserClient<InterceptedService<Channel, InterceptorXToken>>,
+    reconnect_config: ReconnectConfig,
+    geyser_client_opts: TonicGeyserClientOptions,
+    reconnect_endpoint: Option<Endpoint>,
+    reconnect_x_token: Option<AsciiMetadataValue>,
 }
 
 impl GeyserGrpcClient {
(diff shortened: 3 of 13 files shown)


yellowstone-grpc-client/src/lib.rs
```diff
@@ -295,11 +317,11 @@ impl GeyserGrpcBuilder {
 
     // Create client
     fn build(
-        self,
+        &self,
```
lvboudre: builder pattern never take a referential &self for the final build.
it's always you create a builder than collapse it to the built type.

yellowstone-grpc-client/src/lib.rs
```diff
@@ -317,21 +339,28 @@ impl GeyserGrpcBuilder {
...
         Ok(GeyserGrpcClient::new(
             HealthClient::with_interceptor(channel, interceptor),
             geyser,
+            builder,
         ))
     }
 
     pub async fn connect(
-        self,
+        &self,
```
lvboudre: same here.

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,540 @@
...
+                                return Err(status);
+                            }
+                            state.backoff.sleep().await;
+                        }
+                    }
+                }
+            }
+        },
+    )
+}
```
lvboudre: So auto-reconnect feature should be part of the standard `subscribe` method.
It would preferable to create a specialize `Stream` types that hides and implement the auto-reconnect logic if configured.
Having a dedicated `DedupStream` and `AutoReconnect` streams would allow you to create other method, making the code a bit more readable.

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,540 @@
...
+    pub fn call(&self, attempt: u32, status: &Status) {
+        (self.0)(attempt, status);
+    }
+}
+
+impl std::fmt::Debug for ReconnectCallback {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("ReconnectCallback(..)")
+    }
+}
```
lvboudre: I don't understand what the reconnect callback is for?

lvboudre (changes requested):
# Proposed changes 

I propose this:

DedupStream object which implements `Stream` trait and apply the dedup logic you did.

Since we are return `impl Stream` you can return really complex type with lots of nested generics too.


```rust
#[pin_project]
struct DedupStream<GrpcStream> {
  pub(crate) state: DedupState, // this dedup state can recovered by another object on failure.
  #[pin]
  inner: GrpcStream
}

impl<S> Stream< for DedupStream<S> 
  where S: Stream<Item = Result<SubscribeUpdate, Status>> + Send + Unpin + 'static
{
   type Item = <S as Stream>::Item;


   fn poll_next(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Self::Item>> {
       let mut this = self.project();
       loop  {
           let poll = this.inner.poll_next(cx);
           match poll {
              Poll::Ready(result) => {
                   // do dedup or propagate the error
              }
             Poll:Pending => // return pending
       }
    }
}

```

Now for the retry logic, you make make another `Stream` that wraps you dedup stream.


```rust

struct NeedReconnect {
   attempt: usize,
   dedup_state: DedupState,
}

struct ReconnectingFut {
  dedup_state: DedupState,
}

// This is just a prototype but it is essentially a state machine, that each state
// can be represented a concrete type, therefore compiled check.
enum AutoReconnectState<GrpcStream> {
   Connected(DedupStream<GrpcStream>),
   NeedReconnect(NeedReconnect),
   ReconnectingFut(ReconnectingFut) // maybe make it generic?
}


pub struct AutoReconnect<GrpcStream> {
  pub(crate) endpoint: Endpoint, // preconfigured endpoint that builds `Channel`,
  // invariant : Dedup_state can only be `Some` for `inner` is `None`
  state: AutoReconnectState<GrpcStream>,
  reconnect_config: ReconnectionConfig // whatever is the reconnect logic.
}


impl<GrpcStream> Stream for AutoReconnect<GrpcStream> 
   where GrpcStream : // Put all the same condition as before
{
    type  Item = <GrpcStream as Stream>::Item;

    fn poll_next(
        mut self: Pin<&mut Self>,
        cx: &mut Context<'_>,
    ) -> Poll<Option<Self::Item>> {
       let mut this = self.project();
       //  either poll if alive or apply your reconnection state machine.
    }
}
```


Using custom `Stream` types will make the code more elegant instead of cramming all the reconnection logic/state machine inside a loop.

# Dedup window

I didn't see in the case any way to prune the dedup state where you stored all dedup keys. We need a maximum retention window that must be configurable by the customer.

yellowstone-grpc-client/src/lib.rs
Ozodimgba: i changed it to `&self` so the reconnect stream could call `connect()` again but seeing that's the wrong layer to solve it at. will revert build/connect back to self and drop Clone on builder. 

the reconnect stream will hold the endpoint config it needs directly.

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,541 @@
...
+    tonic::{Code, Status},
+    yellowstone_grpc_proto::prelude::{
+        subscribe_update::UpdateOneof, SubscribeRequest, SubscribeUpdate,
+    },
+};
+
+#[derive(Debug, Default)]
+struct DedupTracker {
+    /// Per-slot set of seen message keys
+    seen: BTreeMap<u64, HashSet<DedupKey>>,
```
lvboudre: need some sort of retention window.
Also I using `BtreeMap` for a data structure that will be accessed thousands of time per second is probably not the right choice.

If you need to remember the insert order, simply using `VecDeque` along side the map to know which slot was inserted first.

yellowstone-grpc-client/src/stream.rs
Ozodimgba: it's for observability. when the stream reconnects the user might want to know...log it, bump a metric, pause trading, whatever. tracing/log covers internal logging but the callback lets the consumer react programmatically.

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,541 @@
...
+}
+
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+enum DedupKey {
+    Slot(i32),                  // status
+    Account(Vec<u8>, u64),      // pubkey, write_version
```
lvboudre: write_version is not a good value to use, its different for each RPC.
An account update is unique within a slot using `(Pubkey, Signature)`

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,541 @@
...
+    Transaction(Vec<u8>),       // signature
+    TransactionStatus(Vec<u8>), // signature
+    Entry(u64),                 // index
+    BlockMeta,
+    Block,
```
lvboudre: Instead of `Vec` for Pubkey and Signature, you can using `[u8; 32]` (pubkey) and `[u8; 64]` (signature) instead of doing memory allocation inside `Vec::clone`

lvboudre (changes requested):

yellowstone-grpc-client/src/stream.rs
Ozodimgba: Thanks for this

yellowstone-grpc-client/src/stream.rs
lvboudre: I think we can skip that, in the auto reconnect you can add debug/info log if needed.
Users does want to think about that. They want a seemless, un-interrupted streams of geyser event.

yellowstone-grpc-client/src/stream.rs
Ozodimgba: fair point. BTree does O(log n) per look up...I used it to solve the implicit ordering for cleanup...VecDeque does it cheaper

examples/rust/src/bin/client_with_reconnect.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    env_logger::init();
+    let args = Args::parse();
+
+    let mut client = GeyserGrpcClient::build_from_shared(args.endpoint)?
+        .x_token(args.x_token)?
+        .auto_reconnect(true)
```
leafaar: question: should auto-reconnect be default as true?

good side is, it will help lots of people that will probably not do that/know that.
caveats: people who already implemented that (very few)....

yellowstone-grpc-client/src/lib.rs
```diff
@@ -241,140 +276,111 @@ impl GeyserGrpcClient {
...
-                    "channel cannot be disconnected or full at this point, got error: {e}"
-                ),
-            }
+    ) -> GeyserGrpcClientResult<(SubscribeRequestSink, Streaming<SubscribeUpdate>)> {
+        let (mut subscribe_tx, subscribe_rx) = mpsc::channel(1000);
+        if let Some(request) = request.clone() {
+            subscribe_tx
+                .send(request)
+                .await
+                .map_err(GeyserGrpcClientError::SubscribeSendError)?;
```
lvboudre: This can never error, because subscribe_rx is guarantee to be not full and not disconnect, should be replaced by `expect`.

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,1163 @@
...
+                }
+                if let Some(limit) = max_encoding_message_size {
+                    geyser = geyser.max_encoding_message_size(limit);
+                }
+
+                let (mut subscribe_tx, subscribe_rx) = mpsc::channel(1000);
+                subscribe_tx
+                    .send(request)
+                    .await
+                    .map_err(GeyserGrpcClientError::SubscribeSendError)
```
lvboudre: should never fail, can be replaced by expect.

yellowstone-grpc-client/src/lib.rs
```diff
@@ -54,17 +64,62 @@ impl Interceptor for InterceptorXToken {
 pub enum GeyserGrpcClientError {
     #[error("gRPC status: {0}")]
     TonicStatus(#[from] Status),
+    #[error("Failed to send subscribe request: {0}")]
+    SubscribeSendError(#[from] mpsc::SendError),
```
lvboudre: We don't need this error.

examples/rust/src/bin/client_with_reconnect.rs
Ozodimgba: I think it should but after a lot of testsss

CHANGELOG.md
```diff
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- yellowstone-grpc-client-simple-12.3.0
+
+### Features
+
+- client: auto-reconnect stream for yellowstone-grpc-client ([#717](https://github.com/rpcpool/yellowstone-grpc/pull/717))
+
+### Breaking
+
+- client: `GeyserGrpcClient::new` takes 3 args (added `Option<GeyserGrpcBuilder>`)
+- client: `subscribe_once` requires `F: 'static`
```
leafaar: version mismatch with cargo.toml

examples/rust/src/bin/client_with_reconnect.rs
leafaar: to deploy we need lots of tests, i feel like it will never be used if it doesnt get default true

leafaar: well unless we talk about it

yellowstone-grpc-client/src/stream.rs
```diff
@@ -0,0 +1,1163 @@
...
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+/// Unique identity used to detect duplicate messages per slot.
+enum DedupKey {
+    Slot(i32),                           // status
+    Account([u8; 32], Option<[u8; 64]>), // pubkey, txn_signature
+    Transaction(u64),                    // index
+    TransactionStatus(u64),              // index
+    Entry(u64),                          // index
+    BlockMeta,
+    Block,
```
leafaar: can you also support DeshredTransaction?

yellowstone-grpc-client/src/reconnect.rs
```diff
@@ -0,0 +1,1163 @@
...
+    },
+    tonic::{
+        codec::CompressionEncoding, metadata::AsciiMetadataValue, transport::Endpoint, Code,
+        Status, Streaming,
+    },
+    yellowstone_grpc_proto::{
+        geyser::geyser_client::GeyserClient,
+        prelude::{subscribe_update::UpdateOneof, SubscribeRequest, SubscribeUpdate},
+    },
+};
```
leafaar, line 24: do you think it would make sense renaming this file to `reconnect.rs`?

yellowstone-grpc-client/src/stream.rs
Ozodimgba: Hmm, ok ok but i will like to get past this one and the push a follow up PR

leafaar (changes requested):
I left some comments, looks good, though I think we need to support DeshredTransactions as well, it's another Subscribe method though, so it can be alittle bit annoying

lvboudre (changes requested):

yellowstone-grpc-client/src/lib.rs
```diff
@@ -241,140 +276,111 @@ impl GeyserGrpcClient {
...
-    ///   The stream will yield updates based on the latest request received from the sink.
-    ///
-    /// # Lifecyle and dropping rules
-    ///
-    /// The subscription will remain active until the stream is dropped or the server closes the connection.
-    ///
-    /// # Initial [`SubscribeRequest`]
-    ///
-    /// You have to provide a [`SubscribeRequest`] to the server to start receiving updates.
-    ///
```
lvboudre, line 266: You deleted all the docs

yellowstone-grpc-client/src/reconnect.rs
Ozodimgba: i see the point but the file is cohesive, everything in it serves the subscribe stream lifecycle. renaming to reconnect.rs would miss that dedup is in here too. 

leaning more toward splitting into dedup.rs and reconnect.rs

leafaar: makes more sense

yellowstone-grpc-client/src/stream.rs
Ozodimgba: added the dedup key for DeshredTransaction is in. 

wiring deshred into AutoReconnect needs the connector and some extensions, will do that in a follow-up PR.

will have to discuss, it's not a quick wire-up

yellowstone-grpc-client/src/lib.rs
```diff
@@ -241,140 +291,124 @@ impl GeyserGrpcClient {
...
-    /// * `request`: an optional initial [`SubscribeDeshredRequest`] to send to the server when establishing the subscription.
-    ///   If provided, the server will start sending deshredded updates based on the request immediately after the subscription is established.
-    ///   If not provided, the server will wait for the first request from the sink before sending any updates.
-    ///
     pub async fn subscribe_deshred_with_request(
         &mut self,
         request: Option<SubscribeDeshredRequest>,
-    ) -> GeyserGrpcClientResult<(SubscribeDeshredRequestSink, DeshredStream)> {
+    ) -> GeyserGrpcClientResult<(
+        impl Sink<SubscribeDeshredRequest, Error = mpsc::SendError>,
```
lvboudre: You put back the legacy anonymous type here.

lvboudre (changes requested):
Please put back docstring in the same place, before I do a full review, otherwise I see too many false changes.

yellowstone-grpc-client/src/reconnect.rs
```diff
@@ -0,0 +1,1086 @@
...
+                    .subscribe(tonic_request)
+                    .await
+                    .map_err(GeyserGrpcClientError::from)
+                    .map_err(|e| {
+                        if is_recoverable_client_error(&e) {
+                            ErrorCategory::Retryable(e)
+                        } else {
+                            ErrorCategory::Unrecoverable(e)
+                        }
+                    })?;
```
lvboudre, line 328: two approaches:

With `SubscribeReplayInfo`: Once you connect, before subscribing, you do a `SubscribeReplayInfo`, it tells what is the oldest it can replay, you make sure that slot is < last-slot streamed - MARGIN. 
- If it has this slot -> you can subscribe and replay
- if not you need to reattempt the connection (decrement the reattempt by 1)

>>>> Second approach: support x-min-slot-context in the subscribe route inside the plugin, where if your oldest replay slot is bigger than `x-min-slot-context`, your return a client tonic error code.

lvboudre (changes requested):

lvboudre:
@Ozodimgba 

After discussion you need to change the plugin code aswell, when doing `subscribe` route.

If the user provides a fromSlot `subscribeRequest` in its initial `SubscribeRequest`, you must verify if you have this slot in your replay store, otherwise return an tonic status code, this will trigger your backoff reconnect logic automatically.

yellowstone-grpc-client/src/reconnect.rs
```diff
@@ -306,6 +306,27 @@ impl GrpcConnector for TonicGrpcConnector {
...
+                    if let Some(first_available) = replay_info.into_inner().first_available {
+                        if slot < first_available {
+                            return Err(ErrorCategory::Retryable(
+                                GeyserGrpcClientError::TonicStatus(
+                                    Status::failed_precondition(format!(
+                                        "replay from slot {slot} not available, oldest: {first_available}"
+                                    ))
+                                )
+                            ));
+                        }
```
lvboudre: You should not need to `suybscribe_replay_info` prior to subscribe.
Simply subscribe with fromSlot, and let the server decide if it should return an error, the retry logic will trigger anyway and reattempt elsewhere.

lvboudre (changes requested):

yellowstone-grpc-client/src/reconnect.rs
```diff
@@ -0,0 +1,1087 @@
...
+        }
+
+        Poll::Ready(None)
+    }
+}
+
+/// Returns true if a client error is considered transient and reconnectable.
+fn is_recoverable_client_error(err: &GeyserGrpcClientError) -> bool {
+    match err {
+        GeyserGrpcClientError::TonicStatus(status) => is_recoverable_status_code(status.code()),
```
lvboudre: Can you add an additional check:

```rust
match err {
   TonicStatus(status) => {
      if let Some(source) = status.source() {
         if source.downcast_ref::<tonic::transport::Error>().is_some() {
            return true
         }
      }
      is_recoverable_status_code(status.code())
   }

}
```

lvboudre (changes requested):
The current `geyser_loop` function at line `782` does not allow ops to configure the replay "cushion" and its hardcoded to 10.

We need to support configurable replay buffer capacity size and I would put a default value of 150 ~ 60seconds of slots .
Though we need to talk to ops to make sure it does not cause memory issue.

lvboudre (approved):

Ozodimgba:
> I left some comments, looks good, though I think we need to support DeshredTransactions as well, it's another Subscribe method though, so it can be alittle bit annoying

@leafaar Deshred auto-reconnect needs from_slot added to SubscribeDeshredRequest in the proto for replay to work and without it the reconnect can't tell the server where to resume from. 

I've started the work on a str-404-deshred branch (generalized the connector trait, added Dedupable impl for deshred, marks_slot_complete for SlotProcessed) but it's not ready to ship together.

leafaar (approved):

PR #719, opened 2026-03-26
add precommit hooks for signing + formatting checks

WilfredAlmeida:
- Adds a repository pre-commit hook at pre-commit to enforce signed commits and run cargo formatting checks.
- Blocks commit when `commit.gpgsign` is not enabled (`true`).
- Runs `cargo fmt --all -- --check` at repo root and prints warnings (does not fail commit) when formatting is off.
- If staged files include `yellowstone-grpc-client-nodejs/napi/**`, runs an additional warning-only fmt check in that crate.
- Adds `install-hooks` target in Makefile to set `core.hooksPath` to .githooks.
- Documents setup/behavior in README.md, with minor Markdown formatting normalization in the same file.


Juanito87 (approved):
LGTM

leafaar (approved):

PR #721, opened 2026-03-30
str-421: refactor GeyserGrpcClient types

lvboudre:
# Current issues

- Leaky uncustomizable Generic Interceptor type :  our client type GeyserGrpcClient<I> expect an "interceptor" to inject `x-token`. This generic type is leaked all over our API, though users (customers) cannot customize this type at all, so it pollutes everyone else code because of this and integrate poorly with NAPI sdk which does not handle generic type verywell and forces us to do unsafe downcasting casting.

- Leaky abstraction around subscribe `impl Sink` whose error types is fixed to `futures::channel::error::SendError`.  In other words, because of the error type, we are leaking this underlying channel implementation. It is not abstract at all.
Moreover, we use anonymous types which has more constraint at compile time than having the concrete type in the first place.

- As we are building auto reconnect feature in #717 , it's easier to implement generic connector logic if the `Stream` returned by `subscribe` is not an anonymous type.

# After the PR

- Removed generic interceptor types from `GeyserGrpcClient`
- Replaced `impl Sink<SubscribeRequest, Error = mpsc::SendError> + use<F>` with `SubscribeRequestSink` type.
- Replaced `impl Stream<Item = Result<SubscribeUpdate, Status>> + use<F>` with `GeyserStream` type.
- Applied similar type strategy for deshredded method too.
- `SubscribeRequestSink`now returns `SubscribeRequestSinkError` instead of `mpsc::SendError`.
- Added more documentations on most important method.
- retrofitted `client.rs` in our example folder.
- Removed `GeyserGrpcClientError::SubscribeSendError` variant as no code branch could raise this error.

TL;DR: instead of returning leaky anonymous types like `impl Sink` and `impl Stream` we will return concrete types which will hide the underlying sink or stream. This will give us the flexibility to change the implementation when we want without breaking API in the future while providing a cleaner API with concrete types.

### DIFF
--- a/yellowstone-grpc-client/src/lib.rs
+++ b/yellowstone-grpc-client/src/lib.rs
@@ -5,6 +5,7 @@ use {
         channel::mpsc,
         sink::{Sink, SinkExt},
         stream::Stream,
+        StreamExt,
     },
     std::{path::PathBuf, time::Duration},
     tokio::net::UnixStream,
@@ -53,19 +54,20 @@ impl Interceptor for InterceptorXToken {
 pub enum GeyserGrpcClientError {
     #[error("gRPC status: {0}")]
     TonicStatus(#[from] Status),
-    #[error("Failed to send subscribe request: {0}")]
-    SubscribeSendError(#[from] mpsc::SendError),
 }
 
 pub type GeyserGrpcClientResult<T> = Result<T, GeyserGrpcClientError>;
 
+///
+/// See [`GeyserGrpcBuilder`] for constructing a client with custom options.
+///
 #[derive(Clone)]
-pub struct GeyserGrpcClient<F> {
-    pub health: HealthClient<InterceptedService<Channel, F>>,
-    pub geyser: GeyserClient<InterceptedService<Channel, F>>,
+pub struct GeyserGrpcClient {
+    pub health: HealthClient<InterceptedService<Channel, InterceptorXToken>>,
+    pub geyser: GeyserClient<InterceptedService<Channel, InterceptorXToken>>,
 }
 
-impl GeyserGrpcClient<()> {
+impl GeyserGrpcClient {
     pub fn build_from_shared(
         endpoint: impl Into<Bytes>,
     ) -> GeyserGrpcBuilderResult<GeyserGrpcBuilder> {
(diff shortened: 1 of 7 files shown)


yellowstone-grpc-client/src/lib.rs
```diff
@@ -77,10 +79,145 @@ impl GeyserGrpcClient<()> {
...
+    fn poll_close(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_close_unpin(cx).map_err(Into::into)
+    }
+}
+
+impl GeyserGrpcClient {
+    // TODO: check if we need to make this function `pub(crate` instead as users of the lib should use the builder to construct a client anyway.
```
leafaar, line 217: can we remove this todo?

leafaar (approved):
lgtm

yellowstone-grpc-client/src/lib.rs
```diff
@@ -104,35 +241,70 @@ impl<F: Interceptor> GeyserGrpcClient<F> {
...
         let (mut subscribe_tx, subscribe_rx) = mpsc::unbounded();
         if let Some(request) = request {
-            subscribe_tx
-                .send(request)
-                .await
-                .map_err(GeyserGrpcClientError::SubscribeSendError)?;
+            match subscribe_tx.send(request).await {
+                Ok(_) => (),
+                Err(e) => unreachable!(
+                    "channel cannot be disconnect or full at this point, got error: {e}"
```
WilfredAlmeida: grammar improvement for the error message
`channel cannot be disconnected or is full at this point, got error: {e}`

yellowstone-grpc-client/src/lib.rs
```diff
@@ -142,35 +314,57 @@ impl<F: Interceptor> GeyserGrpcClient<F> {
             .map(|(_sink, stream)| stream)
     }
 
-    // Subscribe Deshred
+    ///
+    /// Subscribe to deshred (only transaction right now).
```
WilfredAlmeida: grammar improvement
`(only transactions right now)`

yellowstone-grpc-client/src/lib.rs
```diff
@@ -142,35 +314,57 @@ impl<F: Interceptor> GeyserGrpcClient<F> {
...
+    ///
+    /// Deshred updates only supports deshredded transaction updates.
+    /// Deshredded update are event happening before any replay, they are not guarantee to be valid or even included in the ledger,
```
WilfredAlmeida: grammar improvements
`Deshredded updates are events happening before any replay, they are not guaranteed to be valid or even to be included in the ledger`

yellowstone-grpc-client/src/lib.rs
```diff
@@ -142,35 +314,57 @@ impl<F: Interceptor> GeyserGrpcClient<F> {
...
         let (mut subscribe_tx, subscribe_rx) = mpsc::unbounded();
         if let Some(request) = request {
-            subscribe_tx
-                .send(request)
-                .await
-                .map_err(GeyserGrpcClientError::SubscribeSendError)?;
+            match subscribe_tx.send(request).await {
+                Ok(_) => (),
+                Err(e) => unreachable!(
+                    "channel cannot be disconnect or full at this point, got error: {e}"
```
WilfredAlmeida: same as above grammar improvements

WilfredAlmeida (approved):
LGTM

PR #724, opened 2026-04-02
Str 428: yellowstone-grpc-napi: use non-generic client types

WilfredAlmeida:
Refactor the GrpcClient internal representation to now use concrete GrpcGeyserClient type that has no generic, thus no need of Boxing.

### DIFF
--- a/yellowstone-grpc-client-nodejs/__tests__/subscribe.test.ts
+++ b/yellowstone-grpc-client-nodejs/__tests__/subscribe.test.ts
@@ -469,7 +469,9 @@ function writeAndWaitCallback(
     };
 
     const timeoutId = setTimeout(() => {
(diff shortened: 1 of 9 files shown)


lvboudre:
For your GrpcClient , you are using tokio Mutex because you acquire lock across await points which is the right mutex to use.
But this is mutex is highly enifficient because it is implemented with a queue.
I see you are using a mutex because all your method for the GrpcClient receives an immutable &self reference.
My question to you: is there any NAPI limitation to use &mut self instead? If no, can you try remove this Mutex ?

lvboudre (approved):
@WilfredAlmeida please explain the Mutex limitation why its there as discuss on slack please since I am less familiar with NAPI. It seems to me we could remove this.

lvboudre (approved):

PR #726, opened 2026-04-10
feat(proto): add slot updates and V2 shred fields to deshred protos

leafaar:


WilfredAlmeida (approved):

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -263,8 +266,9 @@ message SubscribeUpdateDeshred {
     SubscribeUpdateDeshredTransaction deshred_transaction = 2;
     SubscribeUpdatePing ping = 3;
     SubscribeUpdatePong pong = 4;
+    SubscribeUpdateSlot slot = 5;
```
leafaar: adding it so we can know when we start/stop receiving deshred txns. and do reconnection

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -263,8 +266,9 @@ message SubscribeUpdateDeshred {
...
   }
-  google.protobuf.Timestamp created_at = 5;
+  google.protobuf.Timestamp created_at = 6;
```
leafaar: breaking change

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -249,6 +250,8 @@ message SubscribeUpdateDeshredTransactionInfo {
   solana.storage.ConfirmedBlock.Transaction transaction = 3;
   repeated bytes loaded_writable_addresses = 4;
   repeated bytes loaded_readonly_addresses = 5;
+  uint32 completed_data_set_starting_shred_index = 6;
+  uint32 completed_data_set_ending_shred_index_exclusive = 7;
```
lvboudre, line 254: this field name are too long,

I propose `entry_shred_start_index`/`entry_shred_end_index` instead.

yellowstone-grpc-proto/proto/geyser.proto
lvboudre: you cannot do that, this will break back ward compatibility as everyone that do no update their client will try to parse field number 5 as Timestamp instaed of slot.

lvboudre (changes requested):

yellowstone-grpc-proto/proto/geyser.proto
leafaar: it's not entry, a complete data set its a `Vec<Entry>` it would be misleading

examples/rust/src/bin/client.rs
```diff
@@ -643,10 +647,22 @@ impl Action {
...
+                if args.slots {
+                    slots.insert(
+                        "client".to_string(),
+                        SubscribeRequestFilterSlots {
+                            filter_by_commitment: Some(false),
+                            interslot_updates: Some(true),
+                        },
+                    );
+                }
+
```
leafaar, line 660: adding support here because it will make easier to test this feature.

lvboudre (approved):

PR #732, opened 2026-04-22
STR-450: Cuckoo filter account subscriptions

Ozodimgba:
Closes #STR-450


yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -49,11 +49,21 @@ message SubscribeRequest {
...
+  uint64 hash_seed = 5;              // seed for deterministic hashing
+}
+
 message SubscribeRequestFilterAccounts {
   repeated string account = 2;
   repeated string owner = 3;
   repeated SubscribeRequestFilterAccountsFilter filters = 4;
   optional bool nonempty_txn_signature = 5;
+
+  optional CuckooFilter cuckoo_filter = 6;
```
lvboudre: I would rename it cuckoo_account, it's more explicit on waht the cuckoo filter is acting on.

yellowstone-grpc-proto/src/cuckoo/filter.rs
```diff
@@ -0,0 +1,802 @@
...
+        );
+    }
+
+    #[test]
+    fn pubkey_like_data() {
+        // 32-byte arrays like Solana pubkeys
+        let mut filter = CuckooFilter::with_capacity(1000).unwrap();
+
+        for i in 0..100u8 {
+            let pubkey = [i; 32];
```
lvboudre, line 675: I would like to see false positive of <1% unit test of **Random** pubkey like please.

yellowstone-grpc-proto/src/cuckoo/filter.rs
```diff
@@ -0,0 +1,802 @@
...
+        // should still work
+        for i in 0..50u64 {
+            assert!(restored2.contains(&i).unwrap());
+        }
+    }
+
+    // at scale
+
+    #[test]
+    fn hundred_thousand_items() {
```
lvboudre, line 744: I want to make sure that we can support 1% false positive rate with 2 million random account pubkey please.

yellowstone-grpc-proto/src/cuckoo/filter.rs
```diff
@@ -0,0 +1,802 @@
...
+            bucket_count: filter.buckets.len() as u32,
+            entries_per_bucket: ENTRIES_PER_BUCKET as u32,
+            fingerprint_bits: FINGERPRINT_BITS,
+            hash_seed: HASH_SEED,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lvboudre, line 391: I want to see a test for collision,
two items that once hashed and assigned to buckets would collide.

Two different value that collide should be allowed to be inserted.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+}
+
+#[derive(Debug)]
+pub enum CuckooError {
+    CapacityOverflow,
+    TableFull,
+    InvalidState, // corrupted or empty buckets
+}
+
+impl CuckooFilter {
```
lvboudre: Make the Cuckoo Filter generic overt T instead of letting all method generics over T,
otherwise your API does not prevent users from inserted unrelated data to the same filter thus less safe.

yellowstone-grpc-proto/src/cuckoo/filter.rs
lvboudre, line 391: Please add tests to cover the error path

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+    /// Only remove items that were previously inserted. Removing a non-inserted
+    /// item may accidentally remove a different item with the same fingerprint.
+    ///
+    /// # Errors
+    ///
+    /// Returns `CuckooError::InvalidState` if filter has no buckets.
+    pub fn remove<T: Hash>(&mut self, item: &T) -> Result<bool, CuckooError> {
+        if self.buckets.is_empty() {
+            return Err(CuckooError::InvalidState);
+        }
```
lvboudre: removing an item that does not exists should be a no op like any other set or map data structure.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+    ///
+    /// - Returns `Ok(false)`: Item is definitely NOT in the set
+    /// - Returns `Ok(true)`: Item is probably in the set (small false positive chance)
+    ///
+    /// # Errors
+    ///
+    /// Returns `CuckooError::InvalidState` if filter has no buckets.
+    pub fn contains<T: Hash>(&self, item: &T) -> Result<bool, CuckooError> {
+        if self.buckets.is_empty() {
+            return Err(CuckooError::InvalidState);
```
lvboudre: if buckets is empty, it should not return an error, but return false like any other kind of data container.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+#[derive(Debug)]
+pub struct CuckooFilter {
+    buckets: Vec<Bucket>,
+}
+
+#[derive(Debug)]
+pub enum CuckooError {
+    CapacityOverflow,
+    TableFull,
+    InvalidState, // corrupted or empty buckets
```
lvboudre: this shoudl not happen, it's like a "catch all errors"
your API should prevent corruption from happening, if not panic because it's probably a bug.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+///
+/// Stores fingerprints (short hashes) of items rather than items themselves.
+/// Supports insert, lookup, and delete operations.
+#[derive(Debug)]
+pub struct CuckooFilter {
+    buckets: Vec<Bucket>,
+}
+
+#[derive(Debug)]
+pub enum CuckooError {
```
lvboudre: Custom Error type must derive `thiserror:Error` so it generates a `std::error::Error` compatible error type.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,802 @@
...
+const FINGERPRINT_BITS: u32 = 16;
+
+/// Fingerprint type. Must match FINGERPRINT_BITS.
+type Fingerprint = u16; // must match FINGERPRINT_BITS
+
+/// A bucket holds ENTRIES_PER_BUCKET fingerprints. Value 0 means empty slot.
+type Bucket = [Fingerprint; ENTRIES_PER_BUCKET];
+
+/// Hash seed for deterministic behavior. ASCII: "yllwstn!"
+const HASH_SEED: u64 = 0x_796c_6c77_7374_6e21;
```
lvboudre: The HASH used should not be a hardcoded but a be injectred in the constructor.
Many hash-based data container actually have a two generic types: `<T,  S = RandomState>` . We should do the same for your cuckoo filter.

lvboudre (changes requested):
Please address these issues. Ill give a second review afterward.

yellowstone-grpc-proto/src/cuckoo.rs
Ozodimgba: but that requires the user to always be aware about the using the same seed as the server. it is not dummy proof...which is why I hard coded

Ozodimgba: agreed on the generic, the key is that the default S has to be deterministic (not RandomState) or the filter isn't wire-compatible. I am thinking to add a Hash builder as the default so the common path is still safe

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,970 @@
...
+            }
+        }
+
+        Err(CuckooError::TableFull)
+    }
+
+    /// Checks if an item is probably in the filter.
+    ///
+    /// - Returns `Ok(false)`: Item is definitely NOT in the set
+    /// - Returns `Ok(true)`: Item is probably in the set (small false positive chance)
```
lvboudre: docstrings is invalid, you return `bool` not `Ok(bool)` anymore.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,970 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Returns `CuckooError::InvalidState` if filter has no buckets.
```
lvboudre: deprecated docstring.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,970 @@
...
+const FINGERPRINT_BITS: u32 = 16;
+
+/// Fingerprint type. Must match FINGERPRINT_BITS.
+type Fingerprint = u16; // must match FINGERPRINT_BITS
+
+/// A bucket holds ENTRIES_PER_BUCKET fingerprints. Value 0 means empty slot.
+type Bucket = [Fingerprint; ENTRIES_PER_BUCKET];
+
+/// Hash seed for deterministic behavior. ASCII: "yllwstn!"
+const HASH_SEED: u64 = 0x_796c_6c77_7374_6e21;
```
lvboudre: make it public please, and renamed it `DEFAULT_HASH_SEED`.

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,970 @@
...
+//! - `entries_per_bucket`: Slots per bucket (4)
+//! - `fingerprint_bits`: Bits per fingerprint (16)
+//! - `hash_seed`: Seed for deterministic hashing
+//!
+//! This enables cross-language deserialization.
+
+use {
+    crate::geyser::CuckooFilter as ProtoCuckooFilter,
+    std::{
+        hash::{BuildHasher, DefaultHasher, Hash, Hasher},
```
lvboudre: You cannot use `DefaultHasher` since the internal hash algorithm used by it is unstable as described in the standard lib documentation.
You must defined the algorithm you want to use instead.

Add a cargo deps for [siphash standalone crate](https://docs.rs/siphasher/latest/siphasher/)

and use a siphash algorithm.

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -49,11 +49,21 @@ message SubscribeRequest {
   optional uint64 from_slot = 11;
 }
 
+message CuckooFilter {
+  bytes data = 1;                    // bucket data
+  uint32 bucket_count = 2;           // number of buckets
+  uint32 entries_per_bucket = 3;     // slots per bucket (typically 4)
+  uint32 fingerprint_bits = 4;       // fingerprint size (8, 12, or 16)
+  uint64 hash_seed = 5;              // seed for deterministic hashing
```
lvboudre, line 65: To maximise future updates, also include a field that specified the hash algorithm used, create an enum :

`enum CuckooHashAlgorith {
    SIP_HASH = 0
}
`

This will allow us to change the Hasher in the future while staying backward compatible.

yellowstone-grpc-client/src/filter_set.rs
lvboudre, line 1: Everything related to the cuckoo filter should stay in proto crate please.

yellowstone-grpc-client/src/filter_set.rs
```diff
@@ -0,0 +1,248 @@
...
+            CuckooFilter as ProtoCuckooFilter, SubscribeRequest, SubscribeRequestFilterAccounts,
+        },
+    },
+};
+
+/// A HashMap-like wrapper around `CuckooFilter` for safe client-side usage.
+///
+/// Maintains both a `HashSet` (source of truth) and a `CuckooFilter` (for wire transmission).
+/// The `HashSet` ensures exact membership checks and guards against unsafe removes.
+pub struct LocalCuckooMap<T> {
```
lvboudre: drop the "local". just called `CuckooHashMap`

yellowstone-grpc-proto/src/cuckoo/filter.rs
```diff
@@ -0,0 +1,970 @@
...
+        for _ in 0..PROBES {
+            let key: [u8; 32] = probe_rng.gen();
+            if filter.contains(&key) {
+                false_positives += 1;
+            }
+        }
+
+        let fp_rate = false_positives as f64 / PROBES as f64;
+        assert!(
+            fp_rate < 0.01,
```
lvboudre: shouldn't it be `<=` ?

lvboudre (changes requested):

yellowstone-grpc-proto/src/cuckoo.rs
```diff
@@ -0,0 +1,970 @@
...
+    _phantom: PhantomData<fn() -> T>,
+}
+
+#[derive(Debug, Error)]
+pub enum CuckooError {
+    #[error("capacity overflow: requested capacity exceeds maximum")]
+    CapacityOverflow,
+    #[error("cuckoo table full after {} kicks", MAX_KICKS)]
+    TableFull,
+}
```
lvboudre: would drop the capacity overflow error.
It's ambiguous with the other error name TableFull and it's error that can only happen during the initial construction of the cuckoo filter.
You pollute all other API method with this error. For example: insert can never raise CapacityOverflow but you force all client to handle this error case.

Users must know the implementation details that insert will never raise it so they can ignore it -> this is leaky.

It's a code smell to have broad error types.

If some error types can only happen during the constructor, then create a `CuckooFilterInitError` type.

If some error can only happen during the `insert`, then create a `CuckooInsertError`.

lvboudre (changes requested):

yellowstone-grpc-proto/src/cuckoo/map.rs
```diff
@@ -0,0 +1,475 @@
...
+/// map.insert(100).unwrap();
+/// assert!(map.contains(&42));
+/// assert_eq!(map.len(), 2);
+///
+/// map.remove(&42);
+/// assert!(!map.contains(&42));
+/// ```
+///
+/// [`HashSet`]: std::collections::HashSet
+pub struct CuckooMap<T> {
```
lvboudre: rename it to `CompressedFilterSet`

yellowstone-grpc-proto/src/cuckoo/set.rs
```diff
@@ -0,0 +1,475 @@
...
+    pub fn is_empty(&self) -> bool {
+        self.items.is_empty()
+    }
+
+    /// Returns `true` if the map has been mutated since the last call to
+    /// [`take_dirty`] (or since construction).
+    ///
+    /// Use this to check whether the filter needs to be re-sent without
+    /// clearing the flag.
+    pub const fn is_dirty(&self) -> bool {
```
lvboudre, line 234: how can this be const?

yellowstone-grpc-proto/src/cuckoo/set.rs
```diff
@@ -0,0 +1,475 @@
...
+    /// let mut map = CuckooMap::<u64>::with_capacity(100).unwrap();
+    /// assert!(!map.take_dirty());    // fresh map is clean
+    ///
+    /// map.insert(42).unwrap();
+    /// assert!(map.take_dirty());     // mutation flipped it
+    /// assert!(!map.take_dirty());    // and clearing it takes effect
+    /// ```
+    pub const fn take_dirty(&mut self) -> bool {
+        std::mem::replace(&mut self.dirty, false)
+    }
```
lvboudre, line 261: I am not sure about this "diryt" API handling. I would remove it if I were you because it serves no purpose from client perspective: 

If I insert/remove accounts from the filter, I know it's dirty since I just did the action! I still have to `to_proto` and put it inside my `SubscribeRequest`. Looking at `is_dirty` or do `take_dirty` does not add any benefits.

lvboudre (approved):
Please address the following issues.

yellowstone-grpc-proto/src/cuckoo/set.rs
Ozodimgba: const fn just means the body is const-evaluable, doesn't have to be called in a const context. same reason Vec::len is const fn in std even though Vec can't exist at compile time.

yellowstone-grpc-proto/src/cuckoo/set.rs
Ozodimgba: fairs. my thinking was the multi-task case where mutations come from one task and resends from another that user can't observe mutations directly.

but they'd already have a lock around the set in that case, i will remove it

Ozodimgba: on the second thought, I am reconsidering this. 

the "user already knows when they mutated" doesn't hold for the shape we'll actually see, which may be mutation on one task (pool monitor, event consumer), resend on another running cadence-driven.

yes the user could write their own AtomicBool next to the set, but every integrator writing that one line decides for themselves where it lives, what synchronization protects it, when to flip and clear. the flag in the type makes it one canonical answer instead of N slightly different ones. cheap to provide, cheap to remove later.

it serves a purpose for this case

leafaar (approved):

yellowstone-grpc-proto/src/cuckoo/set.rs
lvboudre, line 1: this file should not be called "map.rs" anymore but "set.rs"

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -102,6 +121,7 @@ message SubscribeRequestFilterBlocks {
   optional bool include_transactions = 2;
   optional bool include_accounts = 3;
   optional bool include_entries = 4;
+  optional CuckooFilter cuckoo_accounts_filter = 5;
```
lvboudre: should be renamed `cuckoo_account_include` to be consistent with the API naming convention.

lvboudre (changes requested):

lvboudre (approved):

PR #734, opened 2026-04-24
feat(geyser): dedup deshred transactions in the geyser loop via Forks

leafaar:
A validator receives the same shred over multiple turbine paths: broadcast (direct from the leader) and retransmit (relayed by a peer). Each path independently feeds the deshred pipeline, so notify_deshred_transaction fires more than once with the same (slot, signature). Downstream subscribers only need to see each transaction once.

Dedup state lives inside geyser_loop, which is the single consumer of the message channel, so no Mutex or other synchronization is needed. Duplicate DeshredTransaction messages are discarded the moment they are dequeued, before any bookkeeping, msgid assignment, or broadcast to subscribers.

Slot lifecycle tracking is delegated to
yellowstone_block_machine::forks::Forks so dedup state for an orphaned sibling is released as soon as a competing branch is rooted, instead of waiting for Rooted/Dead. add_slot_with_parent is called on Processed; make_slot_rooted_with_rooted_trace on Finalized (also evicts every ancestor that gets indirectly rooted); mark_slot_as_forked on Dead. truncate_excess_rooted_slots bounds memory by pruning aged-out subtrees.

Forks does not expose an is_forked_slot accessor, so Dead-status slots plus the descendants that mark_slot_as_forked surfaces are recorded in a separate HashSet<Slot>. try_insert rejects post-root slots (via Forks::is_rooted_slot), post-dead slots, and duplicate signatures within the live window. The dead set is pruned alongside seen whenever truncate_excess_rooted_slots ages slots out, so it stays bounded.

Adds yellowstone-block-machine 0.7 and rustc-hash 2.0 to workspace deps. FxHashSet is confined to the one truncate_excess_rooted_slots call site that requires the concrete type; all dedup-owned state uses std HashSet.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -352,6 +352,11 @@ impl SlotMessages {
 
 type BroadcastedMessage = (CommitmentLevel, Arc<Vec<(u64, Message)>>);
 
+/// Messages broadcast on the deshred channel. Deshred is a pre-execution
+/// stream and has no commitment level: each message is emitted exactly
+/// once, when first received from the geyser plugin.
+type DeshredBroadcastedMessage = Arc<Vec<(u64, Message)>>;
+
 enum ReplayedResponse {
     Messages(Vec<(u64, Message)>),
     Lagged(Slot),
@@ -573,6 +578,7 @@ pub struct GrpcService {
     subscribe_id: Arc<AtomicUsize>,
     snapshot_rx: Arc<Mutex<Option<crossbeam_channel::Receiver<Box<Message>>>>>,
     broadcast_tx: broadcast::Sender<BroadcastedMessage>,
+    deshred_broadcast_tx: broadcast::Sender<DeshredBroadcastedMessage>,
     replay_stored_slots_tx: Option<mpsc::Sender<ReplayStoredSlotsRequest>>,
     replay_first_available_slot: Option<Arc<AtomicU64>>,
     debug_clients_tx: Option<mpsc::UnboundedSender<DebugClientMessage>>,
@@ -647,6 +653,8 @@ impl GrpcService {
 
         // Messages to clients combined by commitment
         let (broadcast_tx, _) = broadcast::channel(config.channel_capacity);
+        // Deshred subscribers receive their own commitment-free stream.
+        let (deshred_broadcast_tx, _) = broadcast::channel(config.channel_capacity);
         let (replay_first_available_slot, replay_stored_slots_tx, replay_stored_slots_rx) =
             if config.replay_stored_slots == 0 {
                 (None, None, None)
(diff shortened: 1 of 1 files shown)


lvboudre (approved):

yellowstone-grpc-geyser/src/plugin/deshred_dedup.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    /// Mark `slot` dead. Its entire subtree is evicted from `seen` and
+    /// recorded in `dead` so late arrivals are rejected by `try_insert`.
+    pub fn on_slot_dead(&mut self, slot: Slot) {
+        self.forks
+            .mark_slot_as_forked(slot, &mut self.dropped_scratch);
+        self.dropped_scratch.insert(slot);
+        self.dead.extend(self.dropped_scratch.iter().copied());
```
Ozodimgba: this records as rejected

yellowstone-grpc-geyser/src/plugin/deshred_dedup.rs
```diff
@@ -0,0 +1,225 @@
...
+
+    /// Promote `slot` to rooted. Siblings are marked forked and evicted;
+    /// `slot`'s own dedup set, along with every ancestor that was just
+    /// promoted to rooted indirectly, is also released. Post-root traffic
+    /// is handled by `try_insert`'s `is_rooted_slot` guard.
+    pub fn on_slot_finalized(&mut self, slot: Slot) {
+        let mut indirectly_rooted: HashSet<Slot> = HashSet::new();
+        self.forks.make_slot_rooted_with_rooted_trace(
+            slot,
+            &mut self.dropped_scratch,
```
Ozodimgba: orphaned siblings land here

yellowstone-grpc-geyser/src/plugin/deshred_dedup.rs
```diff
@@ -0,0 +1,225 @@
...
+            &mut indirectly_rooted,
+        );
+        self.dropped_scratch.insert(slot);
+        self.dropped_scratch.extend(indirectly_rooted);
+        self.drain_dropped();
```
Ozodimgba: evicts from `seen` but no equivalent of `self.dead.extend(...)`

Ozodimgba (changes requested):
the finalize path evicts orphaned siblings from seen but doesn't add them to dead the way on_slot_dead does

for example; DeshredTransaction(102, A) forwarded -> Slot(103, Finalized) orphans 102 and drops seen[102] -> late retransmit of (102, A) finds 102 not rooted, not in dead, no seen entry -> forwards a duplicate and re-creates seen[102]. That entry leaks: 102 is no longer in forks, so truncate_excess_rooted_slots never reaches it

this is the exact case dead was added to prevent, the dead path mirrors the eviction into a rejection set, the finalize path doesn't

Per the block-machine README,  https://github.com/rpcpool/yellowstone-block-machine#fork-detection -> "all dead slots are forks, but not all forks are dead slots," so we're guarding the strict subset and leaving the broader case open

add a test asserting late try_insert on an orphaned slot is rejected

PR #745, opened 2026-05-27
Add back `convert` feature with convert_from/convert_to modules

luca992:
Lets downstream depend on the SDK<->protobuf conversion helpers without
pulling in yellowstone-grpc-geyser. Geyser re-exports them so existing
call sites are unchanged.

Closes #710
Closes #723


leafaar:
@luca992 thanks for the PR, could you sign your commit so i can run CI

luca992:
@leafaar done 👍

lvboudre:
@luca992 

Why do proto needs to have all those conversion function? The proto is supposed to only hold dragonsmouth data types used for gRPC that's it. If you need to convert proto data why can't you do it on your own code when you receive it?

luca992:
> @luca992 
> 
> Why do proto needs to have all those conversion function? The proto is supposed to only hold dragonsmouth data types used for gRPC that's it. If you need to convert proto data why can't you do it on your own code when you receive it?

I explained in #710.

Also this is a feature that was removed/moved, which I am re-adding in the same create it was previously available in.

leafaar:
@luca992 the commit still says "Unverified", check this https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

lvboudre:
> > @luca992
> > Why do proto needs to have all those conversion function? The proto is supposed to only hold dragonsmouth data types used for gRPC that's it. If you need to convert proto data why can't you do it on your own code when you receive it?
> 
> I explained in #710.
> 
> Also this is a feature that was removed/moved, which I am re-adding in the same create it was previously available in.

Can't you implement this in your code? Copy paste the conversion function? I don't see what's holding you from doing it.
I agree this was removed from the proto, because from the project structure standpoint it was messy: the conversion functions were only used by the plugin. If you go even more in the past history of the repo, it was in the plugin. The reason for the moved was not explained and decided by one person at the time, not a consensus one. We went back to the original vision and this allow the yellowstone-grpc-proto to be lighter and more focused.

luca992:
> > > @luca992
> > > Why do proto needs to have all those conversion function? The proto is supposed to only hold dragonsmouth data types used for gRPC that's it. If you need to convert proto data why can't you do it on your own code when you receive it?
> > 
> > 
> > I explained in #710.
> > Also this is a feature that was removed/moved, which I am re-adding in the same create it was previously available in.
> 
> Can't you implement this in your code? Copy paste the conversion function? I don't see what's holding you from doing it. I agree this was removed from the proto, because from the project structure standpoint it was messy: the conversion functions were only used by the plugin. If you go even more in the past history of the repo, it was in the plugin. The reason for the moved was not explained and decided by one person at the time, not a consensus one. We went back to the original vision and this allow the yellowstone-grpc-proto to be lighter and more focused.

Obviously I could. But I introduced no new code to the repo. I just moved the existing conversion code that is already exposed in the [yellowstone-grpc-geyser](https://github.com/lvboudre/yellowstone-grpc/tree/master/yellowstone-grpc-geyser) package back into [yellowstone-grpc-proto](https://github.com/lvboudre/yellowstone-grpc/tree/master/yellowstone-grpc-proto) where it was exposed pre v12. And it's just a useful optional utility I think other people use (which they do as I am not the only one who made a ticket). 

And I think it is more "correct" to be in the proto package... as it explicitly deals with transforming to/from the proto types. And the proto package is used by both the client an the server. And having the "convert" feature optionally available on the client is useful.

If you guys maintain a vetted proto <-> solana type conversion anyway, I do not see why exposing it in a lightweight package as an optional feature would be a bad thing?

lvboudre:
As I mentioned, it went from plugin to proto, and then back to plugin, since most of the code is just converting Geyser data into Yellowstone data objects. These conversion functions were always intended to remain private implementation details rather than part of the public API. Downstream users are expected to work with the yellowstone-grpc-proto types directly.

We don’t necessarily want to maintain a public API for conversions between Geyser/Solana crate types, even though users can still implement those conversions themselves if needed. For example, you may want a solana_transaction object, while others may not care about those types at all and instead use their own internal representation.

luca992:
Sure. But you do realize is it's still a public API in the library? Just in a super heavy crate. If you made that actually internal I'd understand. But what you are saying is contradictory.

lvboudre:
@luca992 no, geyser plugin is not a library and not publish on crates io.
If you decide to include it as a git crates in your Rust project, do it at your own risk, it's all private implementation details.
The code is free-software you can fork and adjust to your liking.

luca992:
@lvboudre would you be open to a yellowstone-grpc-convert crate instead of a feature on the proto crate? I ideally just would like to to avoid maintaining the conversion myself completely.

luca992:
Otherwise it seems you do not want this at all. In which case you can just reject the pr.

lvboudre:
@luca992 
Groovie made a package which already contain what you are looking for: https://github.com/grooviegermanikus/yellowstone-grpc-convert.

> would you be open to a yellowstone-grpc-convert crate instead of a feature

Whether its a crate or feature, my concerns is being responsible of a public API about conversion that is supposed to be private implementation details. The PR on its own is fine. But here's some details you need to be aware of:

 We have in our roadmap a refactoring to do that will get "rid" of types generated by the proto and instead use a custom type hierarchy that implements codec for yellowstone-grpc-proto, internally we don't want to use types generated by prost in the future. 
 Once we get there, the convert module you are trying to make public will unlikely be the same or what you desired it to be, thus my hesitation of making it public, otherwise this become a burden on our team to keep available.
 
 You are the second person to request this feature so far, and the first person is building closed-source software.
 So far, I don't see enough traction to get back `convert`.
 
 I will discuss with my teammate about this if its something we would like to support, but so far, like you, I don't want to be the maintainer of the official convert tool. Again it was meant to be private in the first place.

lvboudre:
@luca992 I contacted the other person requesting this feature like you do, to better understand the use case.
Could you tell me why you need to put our types back to solana-crates types and not custom type for example?

luca992:
> @luca992 I contacted the other person requesting this feature like you do, to better understand the use case.
> Could you tell me why you need to put our types back to solana-crates types and not custom type for example?

My main concern is that I want to be able to index live transactions data over grpc streams. And also support backfill of transaction data if the stream goes down by making standard rpc requests for missed transactions. Being able to convert proto data to solana-crate types is useful because I can use the same exact logic for persisting data from both grpc or a standard rpc call when backfilling.

I can deal with different types. But it makes my life easier and my indexing code simpler when I have an accurate conversation.

lvboudre:
We've decided not to go forward with this as an open source library already offer the feature ([here's the link](https://github.com/grooviegermanikus/yellowstone-grpc-convert.)).

We may add it back in the future, but right now we are going through many changes on master to improve performance and ease our maintenance work.

The main reason is that Anza often breaks semver assumptions made by Cargo for solana crates, if we put those convert function back to yellowstone-grpc-proto, then the lib will suffer from the lack of semver convention of solana crates.
It would force us to do many major  release bump without real change asides that dependencies cause breaking during compilation. This problem bite us when we tried to support a "triton-ext" version of the software to build against our public solana forks.

luca992:
Got it thanks for clarifying

PR #747, opened 2026-05-28
Performance optimization

Synestraa:
Added jemalloc as a statically linked library; reason being is that the validator's jemalloc is also statically linked. Due to this, yellowstone-grpc continued to use the regular path to malloc instead of routing to jemalloc's. We've also set a feature flag on jemalloc to allow us to have the .so dynamically loaded, as the requested TLS size when static is too large to load, deferring to dynamically loading on a first-visit basis to each thread.

Changed pre_encoded's definition of Account & Transaction to Vec. It is not cloned directly or indirectly anywhere and it saves (as measured 4 micros) per call from Bytes::from(vec).

Optimized a loop to not use a Vec when not needed in iterating slots, it always has a count of 1 anyways. Now uses an Option.

Changed an internal loop to use a stack allocated Vec. It is not cloned or used in any other way which would make performance degradation.

Updated pre-encoding len calculation to use const values for const sized arrays

Added incoming message sink to remove estimations of expected count of messages


lvboudre (changes requested):
Please add the refactored of the geyser loop previously discussed in order to remove the sleep.
We need to bump geyser plugin only at the patch version.
We also need to write the CHANGELOG.md please.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -772,301 +768,308 @@ impl GrpcService {
...
+                            Err(_) => break,
+                        }
 
-                    // Update blocks info
-                    if let Some(blocks_meta_tx) = &blocks_meta_tx {
-                        if matches!(&message, Message::Slot(_) | Message::BlockMeta(_)) {
-                            let _ = blocks_meta_tx.send(message.clone());
+                        if buffered_messages.len() >= PROCESSED_MESSAGES_MAX {
+                            break;
                         }
```
lvboudre: could be replaced with:
```rust
while let Ok(message) = messages_rx.try_recv() {
 ...
}
```

The safest approach would be to detect `Disconnected` error, in that case it should probably crash.

```rust
Err(Empty) => break,
Err(Disconnected) => exit/crash the function
```

CHANGELOG.md
```diff
@@ -10,6 +10,21 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+## 2026-05-28
```
lvboudre: You need to list 
`- yellowstone-grpc-geyser 13.1.1` so we know exactly what version this message is for.

You can move everything under a `### misc` 
and move the removing of  parallel encoding configuration option under `### Breaking `

lvboudre (approved):
LGTM: left a couple of nitpick to address.

leafaar (approved):

PR #748, opened 2026-05-29
feat: connection-close metrics and client_unresponsive disconnect reason

leafaar:
## What

Adds visibility into *why* a gRPC client connection actually goes away, at two layers.

### `connections_closed_total{reason}` (new, transport layer)
A counter incremented once per TCP connection close, classified from the last socket `io::ErrorKind` observed in `SpyIO`:

- `connection_reset` (peer sent a TCP RST), `broken_pipe` (write to a gone peer), `connection_aborted`, `unexpected_eof`, `timed_out`, `not_connected`
- `clean` (no error: a clean FIN is `Ok(0)`/EOF, and a graceful HTTP/2 GOAWAY also lands here)
- `other` (catch-all, required because `io::ErrorKind` is `#[non_exhaustive]`)

This breaks down what an ungraceful `client_closed` on `grpc_client_disconnects_total` actually was at the TCP level. The kind is captured only on the error branch of the poll methods, so the data path is unaffected (no allocation, no locking, no work on the success path). This is per TCP connection, distinct from the per-subscription `grpc_client_disconnects_total`.

### `client_unresponsive` (new reason on `grpc_client_disconnects_total`)
The 10s heartbeat task already detects a dead "zombie" client (for example one whose filter matches nothing, so the send loop never writes and never notices the peer is gone). It now flags the session before cancelling, so that disconnect is attributed to `client_unresponsive` instead of being mislabeled `server_shutdown`.

## Why not a dedicated keepalive-timeout metric

An HTTP/2 keepalive ping timeout is emitted by hyper as `GOAWAY(NO_ERROR)`, which is byte-for-byte indistinguishable from a clean close at the socket, so it falls into the `clean` bucket. Neither hyper nor tonic exposes a callback for it (confirmed by reading the pinned hyper 1.8.1 / h2 0.4.13 / tonic 0.14.3 source). The heartbeat (`client_unresponsive`) is the practical substitute. The rationale and caveat are documented inline next to the classifier.

## Tests

Unit tests for the close-reason classifier (including the `clean` and `other` buckets) and for the error-capture path on read and write.

## Notes

- Patch bump `yellowstone-grpc-geyser` 13.1.1 -> 13.1.2.
- `cargo clippy --all-targets` clean; full geyser lib test suite green.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -37,7 +37,7 @@ use {
         os::unix::fs::PermissionsExt,
         path::PathBuf,
         sync::{
-            atomic::{AtomicU64, AtomicUsize, Ordering},
+            atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
             Arc, LazyLock, Mutex as StdMutex,
         },
         time::SystemTime,
@@ -1107,6 +1107,7 @@ impl GrpcService {
         debug_client_tx: Option<mpsc::UnboundedSender<DebugClientMessage>>,
         maybe_remote_peer_sk_addr: Option<SocketAddr>,
         cancellation_token: CancellationToken,
+        client_unresponsive: Arc<AtomicBool>,
         task_tracker: TaskTracker,
         subscription_tracker: SubscriptionTracker,
     ) {
@@ -1141,7 +1142,11 @@ impl GrpcService {
                     let _ = stream_tx.try_send(Err(Status::internal(
                         "server is shutting down try again later",
                     )));
-                    session.disconnect_reason = "server_shutdown";
+                    session.disconnect_reason = if client_unresponsive.load(Ordering::Acquire) {
+                        "client_unresponsive"
+                    } else {
+                        "server_shutdown"
+                    };
                     return;
                 }
                 Err(ClientSnapshotReplayError::ClientGrpcConnectionClosed) => {
@@ -1161,7 +1166,11 @@ impl GrpcService {
                 _ = cancellation_token.cancelled() => {
                     info!("client #{id}: cancelled");
                     let _ = stream_tx.try_send(Err(Status::unavailable("server is shutting down try again later")));
-                    session.disconnect_reason = "server_shutdown";
+                    session.disconnect_reason = if client_unresponsive.load(Ordering::Acquire) {
+                        "client_unresponsive"
+                    } else {
+                        "server_shutdown"
+                    };
                     break 'outer;
                 }
                 mut message = client_rx.recv() => {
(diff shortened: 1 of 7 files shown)


yellowstone-grpc-geyser/src/transport.rs
```diff
@@ -197,7 +229,12 @@ where
         cx: &mut std::task::Context<'_>,
         buf: &mut tokio::io::ReadBuf<'_>,
     ) -> std::task::Poll<std::io::Result<()>> {
-        Pin::new(&mut self.get_mut().wrappee).poll_read(cx, buf)
+        let this = self.get_mut();
+        let result = Pin::new(&mut this.wrappee).poll_read(cx, buf);
+        if let std::task::Poll::Ready(Err(e)) = &result {
+            this.last_close_error = Some(e.kind());
+        }
+        result
```
lvboudre, line 237: There are several copy/paste code here.
I suggest you moved the polling of `this.wrappee` into a  private function inside `SpyIO`'s impls:

```rust
impl<IO> SpyIO<IO> {
   pub fn poll_wrappee(&mut self, cx: &mut Context<'_>, buf: &mut ReadBuf<'_>) -> Resullt<(), io::Error> {
      let result = Pin::new(&mut this.wrappee).poll_read(cx, buf);
      if let std::task::Poll::Ready(Err(e)) = &resut {
            this.last_close_error = Some(e.kind());
      }
      result
   }
}
```

then you don't have to copy paste the err variant, each time we `poll_read` the wrappee.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1161,7 +1166,11 @@ impl GrpcService {
                 _ = cancellation_token.cancelled() => {
                     info!("client #{id}: cancelled");
                     let _ = stream_tx.try_send(Err(Status::unavailable("server is shutting down try again later")));
-                    session.disconnect_reason = "server_shutdown";
+                    session.disconnect_reason = if client_unresponsive.load(Ordering::Acquire) {
```
lvboudre, line 1169: I don't think you need Acquire/release semantic here, since this atomic variable does not guard the access of any other data, ie no other memory location must be accessed after a load on `client_unresponsive`,
relaxed should be fine.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1141,7 +1142,11 @@ impl GrpcService {
                     let _ = stream_tx.try_send(Err(Status::internal(
                         "server is shutting down try again later",
                     )));
-                    session.disconnect_reason = "server_shutdown";
+                    session.disconnect_reason = if client_unresponsive.load(Ordering::Acquire) {
+                        "client_unresponsive"
```
lvboudre, line 1146: client_unresponsive is not the right name, its true is different from a graceful server shutdown but the cause of the channel closing between our server and the client may still be caused by us.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1585,6 +1598,7 @@ impl Geyser for GrpcService {
                             // does reject every geyser event thus we never write to the HTTP/2 stream and we never detect that the client TCP connection is closed.
                             // By sending a ping every 10 seconds, we can detect if the client is still alive and if it's not,
                             // we can cancel the client loop.
+                            ping_unresponsive.store(true, Ordering::Release);
```
lvboudre, line 1601: This related to the previous comment before about the invalid naming
`ping_stream_tx.send` may return an error, not because the client has closed the connection, but because our servers has closed for some reason. At this level in tonic framework, its too abstract to know who did what.

lvboudre (changes requested):
It's not a bad idea to have finer metrics, at the same even if we count each kind of io errors that we encounter how ops is supposed to make the link between these error and a specific customer or haproxy? So my question is what monitoring use case does it solve?

lvboudre:
Closing this as I don't get any feedback from you.

PR #750, opened 2026-06-02
geyser: refactor block reconstruction using block-machine

lvboudre:
This PR remove most of the legacy code around rebuilding block and managing rooted tree, it is now entirely done using yellowstone-block-machine crate + `block_reconstruction.rs`.

The PR also make sure the `geyser_loop` preserves the ordering guarantee as the old version. A detailed docstring is present that explain every edge case.

It also removes some Arc clones and vector scans on each commitment promotion.

This PR is part of a bigger initiative to optimize the plugin, and this PR role is to simplify as much as possible the geyser loop to enable more aggressive optimization and code-reshaping.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,13 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+## 2026-06-02
+
+- yellowstone-grpc-geyser 13.2.0
+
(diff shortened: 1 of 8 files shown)


leafaar (approved):
lgtm

Synestraa (approved):
I've pushed a small optimization in geyser_loop. The rest looks good

PR #752, opened 2026-06-03
Comment out draft release setting in goreleaser config

linuskendall:
Remove draft releasing, so release straight to release.


leafaar (commented):
Often we build for testing, or something that is a quick fix etc... Wouldn't it be better mantaining as a draft and when we actually publish we can just go there and remove the draft?

PR #753, opened 2026-06-03
STR-528: Reconnect equivocation guard

Ozodimgba:
Closes #STR-528

### DIFF
--- a/yellowstone-grpc-client/src/dedup.rs
+++ b/yellowstone-grpc-client/src/dedup.rs
@@ -9,12 +9,27 @@ use {
         geyser::SubscribeUpdateDeshred,
         prelude::{
             subscribe_update::UpdateOneof,
-            subscribe_update_deshred::UpdateOneof as DeshredUpdateOneof, SubscribeUpdate,
+            subscribe_update_deshred::UpdateOneof as DeshredUpdateOneof, SubscribeUpdate, SlotStatus
         },
     },
 };
 
 pub(crate) const DEFAULT_SLOT_RETENTION: usize = 250;
+
+// SLOT_CREATED_BANK = 5
+const CREATED_BANK_STATUS: i32 = SlotStatus::SlotCreatedBank as i32;
+
+pub(crate) struct EntryIdentity {
+    pub slot: u64,
+    pub index: u64,
+    pub hash: [u8; 32],
+}
+
+/// An entry whose hash conflicts with the one already recorded for its (slot, index).
+pub(crate) struct EntryDivergence {
+    pub entry: EntryIdentity, // `entry.hash` is the value we just received
+    pub expected: [u8; 32],
+}
 
 /// Wrapper stream that filters out duplicate subscribe updates.
 pub struct DedupStream<S> {
@@ -43,6 +58,15 @@ where
         loop {
             match this.inner.poll_next_unpin(cx) {
                 Poll::Ready(Some(Ok(msg))) => {
+                    if let Some(EntryDivergence { entry, expected }) = this.state.entry_divergence(&msg) {
+                        // different block for the same partial slot on the node we reconnected to.
+                        // FailedPrecondition is NOT in is_recoverable_status_code, so AutoReconnect
+                        // terminates instead of reconnecting.
+                        return Poll::Ready(Some(Err(Status::failed_precondition(format!(
+                            "equivocation: slot {} entry {} hash mismatch across nodes: expected {expected:?}, got {:?}",
+                            entry.slot, entry.index, entry.hash,
+                        )))));
+                    }
                     if this.state.is_duplicate(&msg) {
                         continue;
                     }
@@ -79,6 +103,8 @@ pub struct DedupState {
     slot_processed: HashMap<u64, HashSet<i32>>, // slot -> slot status
     slot_order: VecDeque<u64>,
     slot_retention: usize,
+    // partial slots only: slot -> (entry index -> entry hash). dropped on BlockMeta/prune.
+    entry_hashes: HashMap<u64, HashMap<u64, [u8; 32]>>,
 }
 
 impl Default for DedupState {
@@ -89,12 +115,19 @@ impl Default for DedupState {
             slot_processed: Default::default(),
             slot_order: Default::default(),
             slot_retention: DEFAULT_SLOT_RETENTION,
+            entry_hashes: Default::default(),
         }
     }
 }
 
 pub(crate) trait Dedupable {
     fn extract_key(&self) -> Option<(u64, DedupKey)>;
+
+    /// Identifies an Entry update for partial-slot divergence detection.
+    /// None for non-entry updates.
+    fn entry_identity(&self) -> Option<EntryIdentity> {
+        None
+    }
 }
 
 impl Dedupable for SubscribeUpdate {
@@ -124,6 +157,15 @@ impl Dedupable for SubscribeUpdate {
             UpdateOneof::BlockMeta(m) => Some((m.slot, DedupKey::BlockMeta)),
             UpdateOneof::Block(m) => Some((m.slot, DedupKey::Block(m.slot))),
             UpdateOneof::Ping(_) | UpdateOneof::Pong(_) => None,
+        }
+    }
+
+    fn entry_identity(&self) -> Option<EntryIdentity> {
+        if let Some(UpdateOneof::Entry(m)) = self.update_oneof.as_ref() {
+            let hash = <[u8; 32]>::try_from(m.hash.as_slice()).ok()?;
+            Some(EntryIdentity { slot: m.slot, index: m.index, hash })
+        } else {
+            None
         }
     }
 }
@@ -158,6 +200,10 @@ impl DedupState {
         if let Some((slot, key)) = msg.extract_key() {
             match key {
                 DedupKey::Slot(status) => {
+                    // repeated CreatedBank = rollback signal; never dedup it, record() acts on it
+                    if status == CREATED_BANK_STATUS && self.has_created_bank(slot) {
+                        return false;
+                    }
                     if self
                         .slot_processed
                         .get(&slot)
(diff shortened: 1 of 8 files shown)


yellowstone-grpc-client/src/dedup.rs
```diff
@@ -89,12 +115,19 @@ impl Default for DedupState {
...
         }
     }
 }
 
 pub(crate) trait Dedupable {
     fn extract_key(&self) -> Option<(u64, DedupKey)>;
+
+    /// Identifies an Entry update for partial-slot divergence detection.
+    /// None for non-entry updates.
+    fn entry_identity(&self) -> Option<EntryIdentity> {
```
lvboudre: This is a code smell: you add a function for all trait implementors that only a subset of variant (one actually) can implement. This code smell has a name : too-broad interface type.

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -89,12 +115,19 @@ impl Default for DedupState {
             slot_processed: Default::default(),
             slot_order: Default::default(),
             slot_retention: DEFAULT_SLOT_RETENTION,
+            entry_hashes: Default::default(),
         }
     }
 }
 
 pub(crate) trait Dedupable {
```
lvboudre, line 166: The dedupable trait is overall redundant.  At the end of the day there is no other concrete implement for dedupable then DedupKey, even for testing purposes we don't need another implementation: please remove the Dedupable directly, even your DedupState forces us to use DedupKey.

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -9,13 +9,28 @@ use {
...
+    pub index: u64,
+    pub hash: [u8; 32],
+}
+
+/// An entry whose hash conflicts with the one already recorded for its (slot, index).
+pub(crate) struct EntryDivergence {
+    pub entry: EntryIdentity, // `entry.hash` is the value we just received
+    pub expected: [u8; 32],
+}
+
```
lvboudre: Why not just add more information to DedupKey::Entry ?

lvboudre (changes requested):
You need to revisit the implementation, you introduced several code smells to detect diverging entries.
A simpler solution:

In DedupKey::Entry, add the hash also.

When you insert an entry in your dedup state, first check if the entry index is already registered, if yes -> verify the hash match, else -> error.

The error should bubble up and caused the stream to interrupt.

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -158,6 +200,14 @@ impl DedupState {
...
             match key {
                 DedupKey::Slot(status) => {
+                    // Rollback invalidation depends on receiving CreatedBank. The server only emits
+                    // interslot statuses (CreatedBank etc.) to filters with interslot_updates=true
+                    // (see FilterSlots::get_updates). So this guard is dormant unless the user's slot filter sets interslot_updates.
+
+                    // repeated CreatedBank = rollback signal; never dedup it, record() acts on it
+                    if status == CREATED_BANK_STATUS && self.has_created_bank(slot) {
+                        return false;
+                    }
```
lvboudre: When you receive a second CREATED BANK you need to remove all information about this slot in your dedup state , that is really important.

lvboudre (changes requested):

Ozodimgba:
> You need to revisit the implementation, you introduced several code smells to detect diverging entries. A simpler solution:
> 
> In DedupKey::Entry, add the hash also.
> 
> When you insert an entry in your dedup state, first check if the entry index is already registered, if yes -> verify the hash match, else -> error.
> 
> The error should bubble up and caused the stream to interrupt.

cascading effect of assuming the diverge detection and dedup entry should be different. thanks

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -153,114 +163,102 @@ impl DedupState {
...
     }
 }
+
+impl DedupState {
+    /// Drop all dedup + entry state for a single slot (rollback recovery).
+    pub(crate) fn clear_slot(&mut self, slot: u64) {
+        self.slots.remove(&slot);
+        self.slot_order.retain(|s| *s != slot);
+    }
+}
```
lvboudre: Can you put it back to the same `impl DedupState`

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -153,114 +163,102 @@ impl DedupState {
...
-                DedupKey::Slot(status) => {
-                    if let Some(processed) = self.slot_processed.get_mut(&slot) {
-                        processed.insert(status);
-                    } else {
-                        self.inflight_slot_statuses
-                            .entry(slot)
-                            .or_default()
-                            .insert(status);
+                // second CreatedBank => rollback: wipe the whole slot, then record fresh
+                if status == CREATED_BANK_STATUS && self.has_created_bank(slot) {
```
lvboudre: is is CREATED_BANK Just attempt to clear the slot, no need to check if the slot hsa created bank prior to the clearing.

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -62,31 +68,32 @@ pub(crate) enum DedupKey {
...
     Block(u64),
     DeshredTransaction([u8; 64]), // signature
 }
 
+#[derive(Debug, Default, Clone)]
+struct SlotState {
+    keys: HashSet<DedupKey>,         // inflight_slots[slot]
+    statuses: HashSet<i32>,          // inflight_slot_statuses[slot] + slot_processed[slot]
+    entries: HashMap<u64, [u8; 32]>, // entry_hashes[slot]; index -> hash
+    sealed: bool,                    // "is this slot a key in slot_processed?"
```
lvboudre: I don't like the sealed flag, now everytime you receive new data you must check if not already sealed. It does not properly model the fact a slot goes from "being currently replayed" to "as been replayed".

`SlotState` should only exists as long as its not sealed.
THen once its done, you take ownership of it and in DedupState we keep a list of knowned replayed slot, similar to what we used before.

Using sealed + clearing out entriees + dedup keys is more a patch than currently modeling the new situation we are in.

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -153,114 +163,102 @@ impl DedupState {
...
+            match self.slot_order.pop_front() {
+                Some(slot) => {
+                    self.slots.remove(&slot);
+                }
+                None => break,
             }
         }
     }
 
     #[allow(dead_code)]
```
lvboudre: dead code = can we remove it then?

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -62,31 +68,32 @@ pub(crate) enum DedupKey {
...
 #[derive(Debug, Clone)]
 /// Tracks seen messages per slot so we can filter duplicates during replay after reconnect.
 pub struct DedupState {
-    // slot being processed -> set of message keys seen for that slot
-    inflight_slots: HashMap<u64, HashSet<DedupKey>>, // slot -> message_key
-    // slot being processed -> set of visited slot statuses
-    inflight_slot_statuses: HashMap<u64, HashSet<i32>>, // slot -> slot status
-    /// slot already finalized by BlockMeta -> set of slot statuses seen for that slot
-    slot_processed: HashMap<u64, HashSet<i32>>, // slot -> slot status
+    slots: HashMap<u64, SlotState>,
```
lvboudre: add another field called sealed_slot: HashSet<u64> to keep a compressed set of already processed slot so you don't have to use `sealed` flag variable.

Make sure on bank_created, you remove it from the set too.

lvboudre (changes requested):

yellowstone-grpc-client/src/dedup.rs
Ozodimgba: ok that makes sense

yellowstone-grpc-client/src/dedup.rs
```diff
@@ -17,24 +17,74 @@ use {
...
+            replay: ReplayBuffer::new(),
+        }
     }
 }
 
-impl<S, T> Stream for DedupStream<S>
+impl<S, T> Stream for DedupStream<S, T>
 where
-    T: Dedupable,
+    T: Dedupable + Unpin,
```
lvboudre, line 87: Please remove Dedupable trait, overall it adds noisy abstraction, all this module can only dedup SubscribeUpdate anyway,

lvboudre (changes requested):
I don't see any unit test in `dedup.rs`, I want to see a test case for every 
thing we discussed so far:

- test case around partial slot,
- test case around already sealed slot, but received a second bank created.

yellowstone-grpc-client/src/dedup.rs
Ozodimgba: answered in chat

yellowstone-grpc-client/src/reconnect.rs
```diff
@@ -425,11 +425,20 @@ where
...
+                        if !msg.filters.is_empty()
+                            && msg.filters.iter().all(|f| f == AUTORECONNECT_FILTER_KEY)
+                        {
+                            // matched only our internal key (incl. empty) -> user never asked for this
                             me.inner_stream = Some(stream);
                             continue;
                         }
 
+                        // matched a user filter too -> strip the internal key, forward the rest
+                        msg.filters.retain(|f| f != AUTORECONNECT_FILTER_KEY);
```
lvboudre, line 440: if there is no filter left, it's mean the user did not request it, so you should not forward it

Ozodimgba: if I do this, it will break ping/pong messages. we need to only drop when filters were non-empty before stripping.

lvboudre (approved):

leafaar (approved):

PR #760, opened 2026-06-10
separating messages from deshred channel

Synestraa:
In the previous deshred messages implementation, we would forward deshred messages to geyser_loop and then immediately forward them further to deshred_client_loop. To skip the hop of going from message -> geyser_loop -> deshred_client_loop we just directly send to the deshred_client_loop now


leafaar (approved):
nice! lgtm

lvboudre (approved):

PR #762, opened 2026-06-11
Add token_accounts filter for ATA owner expansion on transaction subscriptions

cbpodd:
## Summary

Adds an optional `token_accounts` field to `SubscribeRequestFilterTransactions`. When set, transaction subscriptions can expand the effective account set of each tx with owners of its pre/post token balances, so `account_include` / `account_exclude` / `account_required` match on token-account owners — not only on the tx's real account keys.

Three modes:

| Mode | Behavior |
|-|-|
| `"none"` (default) | No expansion. Byte-identical to current behavior. |
| `"all"` | Include every parseable owner from the tx's pre OR post token balances. |
| `"balanceChanged"` | Include owners whose token balance changed (compared by `account_index`) or whose token account was closed. |

The proto field is optional, encoded as a string, on field number 30.

## Motivation

Useful for subscribing to "every tx that touches wallet X's tokens" without having to enumerate every ATA wallet X has ever owned, including ATAs created mid-stream. Today this is awkward client-side because:

- New ATAs appear via tx and the client doesn't know about them yet.
- Closed ATAs leave a window where the last balance change is unobservable unless the client was already subscribed.
- Some wallets have hundreds of ATAs; `account_include` size limits make exhaustive enumeration impractical.

`balanceChanged` is the common case ("trades / transfers involving this wallet's tokens"); `all` is for callers who want every tx that even references one of the wallet's token accounts.

## Test plan

- [x] `cargo test -p yellowstone-grpc-geyser --lib` — 91 / 91 pass over 3 consecutive runs (69 stock + 22 new).
- [x] `cargo clippy -p yellowstone-grpc-geyser --all-features --all-targets` — clean.
- [x] `cargo check -p yellowstone-grpc-geyser` — clean.
- [x] When `token_accounts` is omitted / "none", filter output is byte-identical to current behavior.

## Notes

**Why pair pre/post balances by `account_index`.** The `Message.account_keys` array is part of the signed tx payload and is frozen for the lifetime of the tx — execution cannot mutate, reorder, or resize it. LUT-loaded addresses are resolved at tx load time (before any instruction runs) and appended deterministically. So `pre_token_balances[i].account_index = N` and `post_token_balances[j].account_index = N` always refer to the same token-account pubkey, even across open/close.

Account "close" in Solana drops the account's lamports to zero and clears its data; it does **not** remove the pubkey from `account_keys`. A closed ATA simply doesn't appear in `post_token_balances` (no token state to report), but `account_index = N` still points at the same pubkey in the tx. This is what the closed-account sweep relies on: any `account_index` present in pre and absent in post is treated as a balance change.

This makes `account_index` a stable join key and avoids resolving each `TokenBalance.account_index` back to a pubkey on every match.

### DIFF
--- a/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
+++ b/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
@@ -44,6 +44,7 @@ use {
             SubscribeRequestFilterEntry, SubscribeRequestFilterSlots,
             SubscribeRequestFilterTransactions,
         },
+        solana::storage::confirmed_block,
     },
 };
 
@@ -729,10 +730,6 @@ struct FilterTransactionsInner {
     account_include: HashSet<Pubkey>,
     account_exclude: HashSet<Pubkey>,
     account_required: HashSet<Pubkey>,
-    // Read by `FilterTransactions::get_updates` in the follow-up commit
-    // that wires the matching logic; tolerated here so the failing-tests
-    // commit compiles cleanly.
-    #[allow(dead_code)]
     token_accounts: TokenAccountsMode,
 }
 
@@ -833,31 +830,30 @@ impl FilterTransactions {
                     }
                 }
 
+                // Effective account set used by include/exclude/required.
+                // When token_accounts mode is set, owners of pre/post token
+                // balances are included alongside the real account keys.
+                // Built fresh per (filter, tx); see TokenAccountsMode docs.
+                let token_owners = collect_token_account_owners(
+                    inner.token_accounts,
+                    &message.transaction.meta,
+                );
+                let in_effective_set = |pubkey: &Pubkey| -> bool {
+                    message.transaction.account_keys.contains(pubkey)
+                        || token_owners.contains(pubkey)
+                };
+
                 if !inner.account_include.is_empty()
-                    && inner
-                        .account_include
-                        .intersection(&message.transaction.account_keys)
-                        .next()
-                        .is_none()
+                    && !inner.account_include.iter().any(in_effective_set)
                 {
                     return None;
                 }
 
-                if !inner.account_exclude.is_empty()
-                    && inner
-                        .account_exclude
-                        .intersection(&message.transaction.account_keys)
-                        .next()
-                        .is_some()
-                {
+                if inner.account_exclude.iter().any(in_effective_set) {
                     return None;
                 }
 
-                if !inner.account_required.is_empty()
-                    && !inner
-                        .account_required
-                        .is_subset(&message.transaction.account_keys)
-                {
+                if !inner.account_required.iter().all(in_effective_set) {
                     return None;
                 }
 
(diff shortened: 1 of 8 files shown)


pmantica11 (approved):

yellowstone-grpc-proto/proto/geyser.proto
```diff
@@ -114,6 +114,11 @@ message SubscribeRequestFilterTransactions {
   repeated string account_include = 3;
   repeated string account_exclude = 4;
   repeated string account_required = 6;
+  // ATA / token-account expansion control. One of "none" (default),
+  // "balanceChanged", or "all". When set, account_include /
+  // account_exclude / account_required also match against owners of
+  // pre/post token balances on each transaction.
+  optional string token_accounts = 30;
```
lvboudre: An enum would better fit what you are trying model here:

```proto
enum TokenAccountExpansionControlFlag {
  ALL = 0,
  BALANCE_CHANGED = 1
}
```

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -702,6 +730,7 @@ struct FilterTransactionsInner {
     account_include: HashSet<Pubkey>,
     account_exclude: HashSet<Pubkey>,
     account_required: HashSet<Pubkey>,
+    token_accounts: TokenAccountsMode,
```
lvboudre: The "none" option is redundant with Rust's `Option::None` variant.
I would change it for 
`token_accounts: Option<TokenACcountsMode>`

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -694,6 +697,31 @@ enum FilterTransactionsType {
     TransactionStatus,
 }
 
+/// ATA / token-account expansion mode for a transaction filter. When not
+/// `None`, `account_include` / `account_exclude` / `account_required` also
+/// match against owners of pre/post token balances on each transaction.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+enum TokenAccountsMode {
+    #[default]
+    None,
```
lvboudre: Redundant with rust `Option::None` variant.

In the proto file you set it to `optional` which already allow clients to represents the state of "non-applicable".

lvboudre (changes requested):
So TLDR:

- The "none" is redundant with rust optional types.

- I get compilation error when building this branch because you need to update also the `examples/rust/client.rs` + `examples/rust/tx-blocktime.rs` (really easy fix, just add the new fields with None as value).

- can you reformat your code : `cargo +nightly fmt --check` produces many errors that will fail during CI/CD.

yellowstone-grpc-proto/proto/geyser.proto
cbpodd: Done - that makes sens, thanks.

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
cbpodd: Done - makes sense.

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
cbpodd: Done - makes sense.

lvboudre (approved):

PR #769, opened 2026-06-16
ci: enable ts sdk integration tests

WilfredAlmeida:


lvboudre (changes requested):
I see many tests that essenitally: 
build a payload, pass it to mock, await the mock then verify if the return value is the same.

I understand you are trying to test the bridge, but I feel its useless test that won't prevent any bug.
A simple integration test that subscriibe to everything and verify it receives one of each event kind is one test that test >80% of the API surface.

Please remove more test, keep it to the essenital.

.github/workflows/ts-sdk-integration-tests.yml
```diff
@@ -0,0 +1,306 @@
...
+      - '.github/workflows/ts-sdk-integration-tests.yml'
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  NODE_VERSION: '20.18.0'
+
+jobs:
+  # Linux x64 GNU (glibc) - Ubuntu
+  test-linux-x64-gnu:
```
lvboudre, line 23: you should use matrix-os instead of copy/pasting every block.

lvboudre (changes requested):

lvboudre (approved):

PR #770, opened 2026-06-16
blockmeta ordering fixes + prometheus invalid block transactions log

Synestraa:
In 13.2.3, a wrong assumption has been made in the ordering of Block/BlockMeta which caused an unnecessary client patch. 
This patch will correct this assumption.

Additionally, we've added a metric to track blocks which contain an invalid amount of transactions.

Close #STR-560


Ozodimgba (approved):
LGTM

lvboudre (approved):

PR #773, opened 2026-06-18
str-565: custom TLS supports

lvboudre:
This part of the initiative to run a secure yellowstone-grpc server instance without the need of a proxy.
In order to do that, we need to support:

- wildcard SAN certificate
- haproxy's style certification directory : a folder of pem files, each containing a unique cert/key pair.
- multi SNI resolution.

This PR re-implement the TCP and TLS incoming from scratch to circumvent limitation in the tonic framework where as tonic's ServerTlsConfig does not allow us to inject a customer Certificate Resolver, only single identity certificate is supported in tonic which forces us to both reimplement TLS and TCP as tonic coupled the two together.

### DIFF
--- a/yellowstone-grpc-geyser/src/config.rs
+++ b/yellowstone-grpc-geyser/src/config.rs
@@ -152,6 +152,15 @@ pub struct GrpcAddresses {
     pub inner: Vec<GrpcAddress>,
 }
 
+impl IntoIterator for GrpcAddresses {
+    type Item = GrpcAddress;
+    type IntoIter = std::vec::IntoIter<GrpcAddress>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.inner.into_iter()
+    }
+}
+
 impl<'de> Deserialize<'de> for GrpcAddress {
     fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
     where
@@ -259,9 +268,9 @@ impl<'de> Deserialize<'de> for GrpcAddresses {
 #[serde(deny_unknown_fields)]
 pub struct ConfigGrpc {
     /// Multiple addresses of Grpc service.
-    pub address: GrpcAddresses,
+    pub address: Option<GrpcAddresses>,
     /// TLS config
-    pub tls_config: Option<ConfigGrpcServerTls>,
+    pub tls_config: Option<TlsIdentityPair>,
     /// Possible compression options
     #[serde(default)]
     pub compression: ConfigGrpcCompression,
@@ -352,6 +361,32 @@ pub struct ConfigGrpc {
     ///
     #[serde(default)]
     pub traffic_reporting_byte_threhsold: Option<ByteSize>,
+
+    ///
+    /// Optional directory to load TLS certificates for gRPC server. If set, the server will start in TLS mode and load certificates from the provided directory.
+    ///
+    /// This option is mutually exclusive with `tls_config`. If both are set, the server will prioritize `cert_dir`.
+    #[serde(default)]
+    pub cert_dir: Option<PathBuf>,
+
+    #[serde(default)]
+    pub listen: Option<Vec<ListenConfig>>,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(untagged)]
+pub enum GrpcTlsConfig {
+    /// A pair of private-key and cert file paths
+    IdentityPair { identity: TlsIdentityPair },
+    /// HAPROXY-like cert directory with `*.pem` files containing the certs and private keys.
+    CertDir { cert_dir: PathBuf },
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct ListenConfig {
+    pub address: GrpcAddress,
+    pub tls: Option<GrpcTlsConfig>,
 }
 
 impl ConfigGrpc {
@@ -398,7 +433,7 @@ impl ConfigGrpc {
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(deny_unknown_fields)]
-pub struct ConfigGrpcServerTls {
+pub struct TlsIdentityPair {
     pub cert_path: String,
     pub key_path: String,
 }
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -1,7 +1,7 @@
 use {
     crate::{
         block_reconstruction::BlockMachineStorage,
-        config::{ConfigGrpc, GrpcAddress},
+        config::{ConfigGrpc, GrpcAddress, GrpcTlsConfig},
         metered::MeteredLayer,
         metrics::{
             self, incr_grpc_method_call_count, set_subscriber_queue_size,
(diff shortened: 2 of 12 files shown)


Synestraa (approved):

yellowstone-grpc-geyser/src/config.rs
```diff
@@ -259,9 +268,9 @@ impl<'de> Deserialize<'de> for GrpcAddresses {
 #[serde(deny_unknown_fields)]
 pub struct ConfigGrpc {
     /// Multiple addresses of Grpc service.
-    pub address: GrpcAddresses,
+    pub address: Option<GrpcAddresses>,
```
Ozodimgba, line 271: Three fields now encode the same concern (TLS identity for a listener): tls_config, cert_dir, and listen[].tls. The mutual exclusion between the first two is enforced by a runtime log warning in create(), not by the type system. create() has two near-identical listener setup blocks as a consequence. One for address + tls_config/cert_dir, one for listen.

I understand keeping the old fields for backward compat with deployed configs. But currently, any future change to listener setup (configurable keepalive, per-listener rate limits, metrics labels) has to be applied to both code paths, and the compiler won't catch drift.

it would make sense to normalize the legacy fields into Vec<ListenConfig> at deserialization time? Something like a fn listeners(&self) -> Vec<ListenConfig> that migrates address + tls_config/cert_dir into the new shape, so create() only has one block. The old fields stay in the serde layer for compat but nothing downstream reads them directly.

If that's out of scope for this PR, we should note

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -427,8 +436,9 @@ impl Drop for ClientSession {
 }
 
 enum Listener {
-    Tcp(TcpIncoming),
-    Unix(PathBuf, UnixListenerStream), // path needed to remove the socket file on exit
+    Tcp(TritonTcpIncoming),
+    Tls(TlsIncoming),
+    Unix(UnixListenerStream), // path needed to remove the socket file on exit
```
Ozodimgba: is the path that is removed here handled

Ozodimgba: just looked at ```serve_listener```.  The old serve_listener used it to std::fs::remove_file(&path) after shutdown. It seems like new serve_listener is generic over I: Stream and has no visibility into whether it's serving a Unix socket, so the cleanup can't live there anymore.

so stale socket files will persist after shutdown and block rebinding on restart.

the path needs to survive until the server task completes...or if there any thing i am missing either way...i cant explicity see how it is handled

Ozodimgba: maybe a Drop guard on Listener

yellowstone-grpc-tools/src/server/tls.rs
```diff
@@ -0,0 +1,760 @@
...
+
+    fn poll_shutdown(
+        self: Pin<&mut Self>,
+        cx: &mut Context<'_>,
+    ) -> Poll<Result<(), std::io::Error>> {
+        self.project().inner.poll_shutdown(cx)
+    }
+}
+
+pub struct TlsIncoming {
```
Ozodimgba, line 362: Is connection level backpressure handled  somewhere upstream, or should TlsIncoming have a cap? If upstream, maybe a doc comment noting the assumption will help ;).

yellowstone-grpc-tools/src/server/tls.rs
```diff
@@ -0,0 +1,760 @@
...
+    exact: HashMap<String, Arc<CertifiedKey>>,
+    wildcard: Vec<(String, Arc<CertifiedKey>)>,
+}
+
+impl SniResolver {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn add(&mut self, name: &str, certified_key: CertifiedKey) -> io::Result<()> {
```
Ozodimgba, line 51: if `add` took `Arc<CertifiedKey>` instead of wrapping it internally, the caller creates one Arc and passes Arc::clone per name. Zero cert-chain duplication.

currently the caller in `build_sni_resolver_from_cert_dir` calls this in a loop over dns names from the same cert, so certs.clone() runs per name, duplicating the entire cert chain each time.

Ozodimgba: minor nit tho

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -469,6 +479,49 @@ pub struct GrpcService {
...
+    tls_config: &GrpcTlsConfig,
+) -> Result<ServerConfig, TlsConfigLoadError> {
+    match tls_config {
+        GrpcTlsConfig::IdentityPair { identity } => {
+            let cert_pem =
+                std::fs::read(&identity.cert_path).expect("failed to read TLS cert file");
+            let key = std::fs::read(&identity.key_path).expect("failed to read TLS key file");
+
+            let cert_der = CertificateDer::from_pem_slice(&cert_pem).expect("tls cert_path");
+            let key_der = PrivateKeyDer::from_pem_slice(&key).expect("tls key_path");
```
Ozodimgba: the error type already has an Io variant, but the std::fs::read and from_pem_slice calls use expect instead of ?. a bad cert path panics the process instead of returning an error. was that intentional, or should these propagate?

lvboudre: Good catch, Ill remove the expect it should not be there.

yellowstone-grpc-tools/src/server/tls.rs
lvboudre: It's not the job of the TlsIncoming to handle back pressure similar to the previous tonic implementation does.

yellowstone-grpc-tools/src/server/tls.rs
```diff
@@ -0,0 +1,760 @@
...
+    pub fn new(incoming: TcpIncoming, acceptor: TlsAcceptor) -> Self {
+        Self {
+            incoming,
+            acceptor,
+            tasks: JoinSet::new(),
+        }
+    }
+}
+
+impl Stream for TlsIncoming {
```
Ozodimgba, line 378: we can log insteadof eprint, no?...also, the poll_join_next drain logic is duplicated between the top of the loop and the Pending arm, we could extract to a helper."

yellowstone-grpc-tools/src/server/tls.rs
lvboudre: this is done only at boot + cloning it is quite cheap.
The API should not force callsite to use Arc because of how `build_sni_resolver_from_cert_dir` is implemented.

yellowstone-grpc-geyser/src/config.rs
lvboudre: Eventually I want to get rid of `address`, I cannot convert it to an array otherwise it will fail deserialization process for existing rpc nodes.
Once of our fleets is updated we will get rid of this.
Also the `address` is stuck with `cert_dir` and `tls_identity` which is also deprecated as it force all address to use the same TLS setup.

lvboudre: Btw the code is already unified both endup in the same Listener object, its just the ServerConfig build that changes.
keepalive/ratelimits etc is done way later in the code over the Listener variants.

yellowstone-grpc-geyser/src/grpc.rs
lvboudre: ill rollback this then

PR #776, opened 2026-06-19
Str 566

lvboudre:
This PR refactor metered bandwidth by moving it to the yellowstone-grpc-tools crates and making it flexible to add custom hooks to measure and monitor bandwidth usage based on custom rules.

Prior to this PR, bandwidth metered was done in the `metered.rs` and coupled with prometheus metrics tracking.
Now, we have `PrometheusMeteredBandwidthHooks` which is responsible to track prometheus metrics without the need to implement all the tonic related objects.

Laslty, this PR introduce rate-limited transport types to limit the number of concurrent TCP connection per remote IP address.

All these are necessary to move to a proxyless setup where we gonna need to implement custom rate limiting and custom metered billing down the road, this is why the metered refactor was necessary and the first step towards this goal.

## Remove

I remove the SpyIncoming which removes one metrics from prometheus : traffic sent per ip address.
AFAIK this metrics was ignored by ops because of the high cardinality it caused.


leafaar (approved):

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
```diff
@@ -0,0 +1,372 @@
...
+
+///
+/// Shared state for tracking active connections per peer IP across all incoming streams and connections. The key is the peer IP and the value is the number of active connections from that IP.
+///
+#[derive(Debug, Clone, Default)]
+pub struct SharedRateLimitTable {
+    inner: Arc<Mutex<HashMap<IpAddr, u64>>>,
+}
+
+pub const DEFAULT_MAX_IP_CONNCUR: u64 = 250;
```
Ozodimgba, line 53: `DEFAULT_MAX_IP_CONNCUR = 250` here and `ConfigGrpc::default_max_ip_conncur() -> 250` in config.rs are two independent sources of truth for the same operational default.

easy for one to get stale

yellowstone-grpc-tools/src/server/tonic/metered.rs
```diff
@@ -0,0 +1,565 @@
...
+}
+
+#[pinned_drop]
+impl<B, MH> PinnedDrop for MeteredBandwidthBody<B, MH>
+where
+    MH: MeteredBandwidthHooks,
+{
+    fn drop(self: Pin<&mut Self>) {
+        // This Drop impl is required to satisfy the !Unpin requirement of the
+        // pin projection in [`MeteredBandwidthBody`], but it should never actually be called
```
Ozodimgba: The comment says "should never actually be called" but Drop always runs (client disconnect, stream end, error), and the body does essential work: flushing cumulative_bytes for the final partial batch. If someone trusts this comment and removes the body, final-batch metrics are silently lost.

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
```diff
@@ -0,0 +1,372 @@
...
+    CB: RatelimitedCallbacks + Unpin + Send + Sync + 'static,
+{
+    type Item = std::io::Result<RateLimitedIO<IO>>;
+    /// Polls for the next accepted connection and wraps it as [`RateLimitedIO`].
+    fn poll_next(
+        mut self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Option<Self::Item>> {
+        let max_ip_conncur = self.max_ip_conncur;
+        let active_conn_map = Arc::clone(&self.active_conn_map);
```
Ozodimgba: Arc::clone and the max_ip_conncur copy both run before poll_next_unpin, so every Pending wakeup pays an atomic increment + decrement for nothing.

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
```diff
@@ -0,0 +1,372 @@
...
+    /// Shuts down the wrapped IO.
+    fn poll_shutdown(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<std::io::Result<()>> {
+        Pin::new(&mut self.get_mut().wrappee).poll_shutdown(cx)
+    }
+}
+
+impl<IO> Drop for RateLimitedIO<IO> {
```
Ozodimgba, line 149: `or_insert(0)` and if `*count > 0` are dead code because `try_new` guarantees the entry exists at >= 1 for any successfully constructed RateLimitedIO, and Drop only runs on successfully constructed values.

maybe unconditional decrement + debug_assert! on entry existence?...makes the invariant visible instead of hiding it behind silent fallbacks.

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
```diff
@@ -0,0 +1,372 @@
...
+    ) -> std::task::Poll<std::io::Result<()>> {
+        Pin::new(&mut self.get_mut().wrappee).poll_read(cx, buf)
+    }
+}
+
+impl<IO> AsyncWrite for RateLimitedIO<IO>
+where
+    IO: AsyncWrite + Unpin,
+{
+    /// Writes to the wrapped IO and updates buffered traffic counters.
```
Ozodimgba: I think this is a stale doc from SpyIO. Body is a pure passthrough, no traffic counters here.

yellowstone-grpc-tools/src/server/tonic/metered.rs
lvboudre: its stale comment ill delete it.

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
lvboudre: copy/paste issue

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
lvboudre: You are right, Ill do expect to be really strict and crash fast otherwise.

yellowstone-grpc-tools/src/server/tonic/metered.rs
```diff
@@ -0,0 +1,565 @@
...
+
+    /// Builds per-request hooks used to meter the response body.
+    ///
+    /// # Parameters
+    /// - `header_map`: Incoming request headers.
+    /// - `uri_path`: Request URI path (for example, `/geyser.Geyser/Subscribe`).
+    ///
+    /// # Returns
+    /// A hooks instance attached to the response body wrapper.
+    fn build_hooks(&self, header_map: &HeaderMap, uri_path: &str) -> Self::Hooks;
```
Ozodimgba, line 90: the old `TRAFFIC_SENT_PER_REMOTE_IP` is gone but nothing replaces it. build_hooks only gets (header_map, uri_path), on_emit_bytes only gets (byte_count, Instant, SystemTime). 

no peer ip anywhere in the new trait. intentional?

yellowstone-grpc-tools/src/server/tonic/ratelimit/transport.rs
lvboudre: I personnally think 250 is way too much as default and config.rs is not part of this library.
I try to write this yellowstone-grpc-tools completely separatly from yellowstone-grpc-geyser.
All code inside `yellowstone-grpc-tools` are agnostic to the business domain,  config.rs in geyser is due to concerns from linus.

yellowstone-grpc-tools/src/server/tonic/metered.rs
lvboudre: yes, we are removing it, ops were not using it, and this metric had too high cardinality.

Ozodimgba (approved):

Synestraa (approved):

PR #793, opened 2026-06-24
Block reconstruction message batching

Synestraa:
Batching messages which we are forwarding to the block reconstruction loop allows us to both reduce the latency of processed messages as well as the latency and throughput of block reconstruction


lvboudre (approved):

leafaar (approved):
amazing!

Issue #792, opened 2026-06-24
From Slot subscription now getting two messages instead of one


grooviegermanikus:
Hey.

I've experienced a (new-to-me) pattern when subscribing to _slots_, which now gives **two** messages for confirmed slots where older versions of _yellowstone-grpc_ gave only **one**.

I could imagine this is intended behavior, but I didn't find any reference to it.

### Example
```
[2026-06-24T08:39:24Z INFO  client] slot (client) at 1782290364.964031: {"deadError":null,"parent":null,"slot":428556486,"status":"SLOT_CONFIRMED"}
[2026-06-24T08:39:24Z INFO  client] slot (client) at 1782290364.964157: {"deadError":null,"parent":428556485,"slot":428556486,"status":"SLOT_CONFIRMED"}
```

### Version and Reproduction
tag: client-v12.1.0

> cargo run --bin client -- --endpoint https://something.com --x-token fooo-barr subscribe --slots

> [2026-06-25T07:02:48Z INFO  client] response: GetVersionResponse { version: "{\"version\":{\"package\":\"yellowstone-grpc-geyser\",\"version\":\"13.2.5-triton-ext\",\"proto\":\"12.4.0\",\"solana\":\"4.0.0\",\"git\":\"geyser-v13.2.1-13-ge2902ab\",\"rustc\":\"1.93.1\",\"buildts\":\"2026-06-17T19:17:02.053949367Z\"},\"extra\":{\"hostname\":\"ZZZZZ\"}}" }

leafaar:
Interesting! Could you run an unary call with getVersion so we know what is the version of the server?

Ozodimgba:
I'm assuming that your're on processed commitment level no?

grooviegermanikus:
no, not only processed. for slots I'm looking at all commitment levels. the issue happens only for slots at level=CONFIRMED.
in old versions that data point did not carry a parent_slot; but now there is one msg with parent_slot=null and one !=null

> yellowstone-grpc/target/debug/client ... subscribe --slots

keirsalterego:
@grooviegermanikus is this already resolved ??

grooviegermanikus:
hey No Not really.
it got even worse.. saw  a third case!

so from my memory there are now three cases:
a- slot confirmed with parent=null
b- slot confirmed twice with parent=null then parent!=null
c- slot confirmed with parent!=null

lvboudre:
[...]

PR #794, opened 2026-06-25
fix: double slot message

Ozodimgba:
#### Problem

Processed subscribers receive duplicate Confirmed/Finalized slot messages for the same slot with different parent fields. One has parent: null, the other has the real parent. [Duplicate Slot Issue](https://github.com/rpcpool/yellowstone-grpc/issues/792)

#### Fix

All commitment statuses should flow through the block machine, never broadcast directly from geyser_loop. The existing code comment already states this ("we must go through the block machine to make sure users see block content before any commitment update"). Current code breaks this by letting commitment statuses into processed_messages.

Redirect commitment statuses (Processed, Confirmed, Finalized) to block_reconstruction_tx in geyser_loop, same pattern as BlockMeta.

### DIFF
--- a/yellowstone-grpc-geyser/src/grpc.rs
+++ b/yellowstone-grpc-geyser/src/grpc.rs
@@ -994,9 +994,21 @@ impl GrpcService {
                     };
                     metrics::message_queue_size_dec();
 
-                    if matches!(&message, Message::BlockMeta(_)) {
-                        let _ = block_reconstruction_tx.send(message);
-                        continue;
+                    match &message {
+                        Message::BlockMeta(_) => {
+                            blockmeta_detected.push(message);
+                            continue;
+                        }
+                        Message::Slot(s) if matches!(s.status,
+                            SlotStatus::Processed |
+                            SlotStatus::Confirmed |
+                            SlotStatus::Finalized
+                        ) => {
+                            metrics::update_slot_plugin_status(s.status, s.slot);
+                            blockmeta_detected.push(message);
+                            continue;
+                        }
+                        _ => {}
                     }
 
                     processed_messages.push(message);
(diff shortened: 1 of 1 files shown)


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -994,19 +994,43 @@ impl GrpcService {
...
+                            blockmeta_detected.push(message);
+                            continue;
+                        }
+                        Message::Slot(s) if matches!(s.status,
+                            SlotStatus::Processed |
+                            SlotStatus::Confirmed |
+                            SlotStatus::Finalized
+                        ) => {
+                            metrics::update_slot_plugin_status(s.status, s.slot);
+                            blockmeta_detected.push(message);
```
lvboudre, line 1008: We should probably rename `blockmeta_detected` for something else.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -994,19 +994,43 @@ impl GrpcService {
...
+                            Message::BlockMeta(_) => {
+                                blockmeta_detected.push(message);
+                                continue;
+                            }
+                            Message::Slot(s) if matches!(s.status,
+                                SlotStatus::Processed |
+                                SlotStatus::Confirmed |
+                                SlotStatus::Finalized
+                            ) => {
+                                metrics::update_slot_plugin_status(s.status, s.slot);
```
lvboudre, line 1029: you call the metric twice.

lvboudre (changes requested):

yellowstone-grpc-geyser/src/grpc.rs
Ozodimgba: yeah not duplicated. a commitment status enters through either the first recv or the try_recv drain, never both.
each path calls the metric once. 

but it gets me thinking the reason it appears in two places is because the loop has two entrypoints...that is not good

leafaar (approved):

lvboudre:
Already merged

PR #795, opened 2026-06-26
Str 567: In-plugin auth resolution

lvboudre:
This PR introduce in-plugin Authentication resolution by adding an optional authentication tonic-layer as the first request receptor of any incoming gRPC request.

# yellowstone-grpc-tools

Added `server::tonic::auth` module which exposes `service.rs` file that contains the generic handling of authentication and expose a public trait that custom authenticator must implement called `SharedAuthenticator`.

**note**: This entire crate is not aware of geyser at-all and just expose generic and cross-domain gRPC/tonic tools.

# yellowstone-grpc-geyser

Added `auth.rs` module that contains multiple implementation of `SharedAuthenticator` trait:
- `TrustedMetadata` : closest to what we have here where we assume the subscription-id resolution has been made prior to calling the service (probably via a proxy).
- `HttpSubscriptionRepository` : use a remote REST service to resolve a pair of `host` + `x-token`, if return 404 -> UNAUTHORIZED is raised.
- `FileSubscriptionRepository` : use a static file with subscription mappings, mostly used for testing purposes.

More details was added to the `README.md` with example.


README.md
lvboudre, line 1: @gritsly @linuskendall here's the operator changes of this PR.

yellowstone-grpc-tools/src/server/tonic/auth/service.rs
lvboudre, line 1: @leafaar @Ozodimgba @Synestraa this is the most important part of the PR.

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1534,19 +1545,81 @@ impl GrpcService {
...
-            .layer(interceptor::InterceptorLayer::new(XTokenInterceptor {
-                x_token,
-            }))
-            .add_service(health_service)
-            .add_service(service)
-            .serve_with_incoming_shutdown(incoming, shutdown.cancelled())
-            .await
-            .map_err(Into::into)
+
+        enum AuthLayerChoice {
```
lvboudre, line 1549: @leafaar @Ozodimgba @Synestraa this is where the entire optional authentication layer is stitched.

yellowstone-grpc-geyser/src/auth.rs
lvboudre, line 1: @leafaar @Ozodimgba @Synestraa 
This file implements new trait from the yellowstone-grpc-tools.

README.md
```diff
@@ -22,6 +22,234 @@ solana-validator --geyser-plugin-config yellowstone-grpc-geyser/config.json
...
+GET http://127.0.0.1:8080/?host=api.example.com&token=test
+```
+
+Example `200 OK` response body:
+
+```json
+{
+    "subscription_id": "test-sub",
+    "rate_limits": {
+         "requests_per_second": 100
```
linuskendall: this one doesn't match the format further down

Synestraa (approved):

yellowstone-grpc-geyser/src/auth.rs
```diff
@@ -0,0 +1,622 @@
...
+    http::{request::Parts, HeaderMap, HeaderName, HeaderValue, StatusCode},
+    moka::future::{Cache, CacheBuilder},
+    reqwest::{IntoUrl, Url},
+    std::{collections::HashMap, convert::Infallible, str::FromStr, sync::Arc},
+    yellowstone_grpc_tools::server::tonic::auth::service::SharedAuthenticator,
+};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct SubscriptionKey {
+    pub xtoken: String,
```
leafaar: why not `x_token` to follow the pattern?

leafaar (approved):

yellowstone-grpc-tools/src/server/tonic/auth/service.rs
```diff
@@ -0,0 +1,811 @@
...
+
+///
+/// Tower service layer that applies a shared authenticator to incoming requests, returning 401 Unauthorized for requests that fail authentication and forwarding authenticated requests to the inner service.
+///
+/// # Generic Parameters
+/// - `Auth`: The type of the shared authenticator. Must implement [`SharedAuthenticator`].
+/// - `S`: The type of the inner service to which authenticated requests will be forwarded.
+///
+pub struct AuthGuardedService<S, Auth> {
+    inner: Arc<Mutex<S>>,
```
Ozodimgba, line 75: I'm not fully sure about this but uh Mutex::lock() inside Future::poll is a blocking call, so if two connections contend on it at the same time one tokio worker thread just sits there frozen, and all connection clones share this same mutex via Arc::clone.

during review i found out that `tower` has a standard pattern for this: call poll_ready on &mut self.inner in poll_ready, then `mem::replace` in call to move the ready service into the future, so it owns S directly with no Arc<Mutex> (requires S: Clone which holds since GeyserServer wraps Arc).

yellowstone-grpc-geyser/src/auth.rs
```diff
@@ -0,0 +1,622 @@
...
+            ]),
+        }
+    }
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct SubscriptionInfo {
+    pub subscription_id: String,
+    #[serde(alias = "rate_limits")]
+    pub ratelimits: Option<SubscriptionRateLimits>,
```
Ozodimgba, line 222: nit: SubscriptionInfo holds a hashmap inside ratelimits, so every clone deep-copies all the String keys + the map, and there are like four call sites that do this across the different auth paths

The data is immutable after construction so maybe a Arc<SubscriptionInfo>, which  would make clone a refcount bump, we could set `AuthExtension = Arc<SubscriptionInfo>` and store Arc in the cache and constant repository.

yellowstone-grpc-geyser/src/auth.rs
```diff
@@ -0,0 +1,622 @@
...
+        host: &str,
+        forwarded_headers: HeaderMap<HeaderValue>,
+    ) -> BoxFuture<'static, Result<Option<SubscriptionInfo>, HttpTritonSubscriptionRepositoryError>>
+    {
+        // To share cache across task, clone it, its cheap.
+        let maybe_cache = self.cache.clone();
+        let mut url = self.url.clone();
+        url.query_pairs_mut().append_pair("host", host);
+
+        let mut caching_key = String::new();
```
Ozodimgba: The caching key concatenates host, header names and values with no delimiter between them, so in theory different inputs can produce the same key but with the current default config where forwarded_headers is just ["x-token"], this is actually hard to hit in practice since the header name is fixed, but it is still structurally problematic. 

Also minor: shouldn't "forwarder" in the expect message should be "forwarded"?

yellowstone-grpc-geyser/src/auth.rs
```diff
@@ -0,0 +1,622 @@
...
+                    if let Some(subscription_info) = maybe {
+                        Ok(Some(subscription_info))
+                    } else {
+                        let maybe_err = err_cell.lock().unwrap().take();
+                        if let Some(err) = maybe_err {
+                            Err(err)
+                        } else {
+                            Ok(None)
+                        }
+                    }
```
Ozodimgba: uh so there might be a subtle issue with the error smuggling pattern here.. each call creates its own err_cell, but moka's optionally_get_with coalesces concurrent requests for the same key.

So if two requests hit the same cache key at the same time and the winning requests init future errors, it writes to its own `err_cell` and returns `None` to moka, then moka gives None to both callers. 

The winning caller checks its err_cell, finds the error, returns Err...but the losing caller's err_cell was never written to, so it falls through to Ok(None) and treats a network failure as "subscription not found" which means a 401 for a potentially legitimate user. 

I think moka's try_get_with propagates errors to all coalesced callers natively but it doesn't have the "optionally" behavior so might need some thought on how to handle 404s.

yellowstone-grpc-geyser/src/auth.rs
lvboudre: the x-token value is unique how can two inputs produce the same key?

yellowstone-grpc-geyser/src/auth.rs
lvboudre: Good catch, ill wrap this in a loop and add atomic flags to determine if the call has actaully been made.

yellowstone-grpc-geyser/src/auth.rs
Ozodimgba: I was thinking more about the general case if forwarded_headers ever gets multiple entries or variable-length header names, but fair point I see it's not a practical issue. The \0 separator is cheap insurance but I wouldn't block on it.

yellowstone-grpc-tools/src/server/tonic/auth/service.rs
lvboudre: Yes you are right about poll_ready, at the same you probably don't have the full picture :

`call` should always follow a `poll_ready` because `poll_ready` may acquire some resources that must be evenutally consumed. This is the case here, we consume a permit and we must spend it.

The issue with using clone is that it does not guarantee it will clone the exact same instance of the underlying generic `Auth` since `Clone` trait does not impose constraint to be a shallow clone.

Later in the AuthFuture I must `poll_ready` the actual inner service which requires `&mut` of `S`, I must be sure its the same instance from between the time I acquire the permit and the time I actually call the service otherwise `poll_ready`/`call` semantic is useless.

>  then mem::replace in call to move the ready service into the future, so it owns S directly with no Arc

And if the underlying Future that took ownership of the `S` previously `mem:replace`ed, is canceled I will lose the service.

Many tower service uses `Clone` too, but there is an implicit constraint here not enforced by the framework where `Service` you implement should hold handles so Clone is shallow, otherwise the `poll_ready`/`call` semantic would not work.

> Mutex::lock() inside Future::poll is a blocking call so if two connections contend on it at the same time one tokio worker thread just sits there frozen.

is not an anti-pattern and alot of Future out there uses mutex under the hood. The only place its dangerous to use mutex in async code is across `await` points which the `poll` isn't. 

In term of performance the lock is only held during the "resolution" of a request before returning a response (mostly streams). Most of user activity lives in the stream handling which is after the lock has been released. Mutex in rust can be called tens of thousands of time per second, far more than the current RPS we are serving in production.
Is there a particular case or thing I am not seeing that Mutex might cause issue @Ozodimgba ?

yellowstone-grpc-geyser/src/auth.rs
lvboudre: `SubscriptionInfo` will be really small, this type will be available through the `Request::extension`  which feels much cleaner if we return `SubscriptionInfo` and not `Arc<SubscriptionInfo>`, I feel its leaky and unintuitive.
Maybe the internal map could be an Arc?

yellowstone-grpc-tools/src/server/tonic/auth/service.rs
Ozodimgba: yeah ok fair, it is true that the lock isn't held across await points and the critical section is just poll_ready + call so contention is realistically negligible... 

I was originally thinking about like a burst scenario where a bunch of connections auth futures all resolve at the same time and they all need the lock sequentially, but at current RPS that's not a real concern. 

The thing I was less sure about is the cancellation point though, with the mem::replace pattern the cancelled future drops a clone not the original service since self.inner gets the clone and the ready instance goes into the future...but your point about Clone not guaranteeing shallow clone for a generic layer is fair, that's a real constraint I didn't think about. I'm good with this. Thanks for explaning

yellowstone-grpc-geyser/src/auth.rs
Ozodimgba: yeah that makes sense, looking at it Arc<SubscriptionInfo> in extensions is leaky. 

I agree with Arc on the internal ratelimits, it is a good middle ground...that way SubscriptionInfo::clone() is just one String + a refcount bump and then consumers don't have to care about the wrapping.

yellowstone-grpc-geyser/src/auth.rs
lvboudre: I don't understand how adding separator prevent this issue, can you ellaborate?

Ozodimgba: so the scenario is when forwarded_headers has multiple entries.. say `["a", "b"]`, without delimiters:
request 1: header `"a"` value `"xb",` header `"b"` value `"y"` -> key contains `"axbby"`
request 2: header `"a"` value `"x",` header `"b"` value `"by"` -> key contains `"axbby"`

so same key but different values. I'm saying that the boundary between one header's value and the next header's name is ambiguous without a separator. With \0 between components the boundaries are fixed. 

but I think this only matters if forwarded_headers ever gets multiple entries, with just `["x-token"]` there's only one header so the value/name boundary can't collide.

Ozodimgba (approved):

yellowstone-grpc-geyser/src/auth.rs
lvboudre: After trying to implement it, we cannot use Arc for serde macros, since its a really small map, Ill go with cloning.
If the map every grow to large we can then do manual serde.

PR #803, opened 2026-07-02
Str 569: Supports optional gRPC method ratelimiting

lvboudre:
Added supported for method ratelimiting inside the plugin. All ratelimiting is implemented with token-bucket so ratelimiting is a rolling window.

This PR is part of an effort to run the entire plugin without proxy.

Ignore yellowstone-grpc-e2e-tests tool refactor.


yellowstone-grpc-geyser/src/ratelimit.rs
lvboudre, line 1: @leafaar @Synestraa @Ozodimgba  this is an important file to review.

yellowstone-grpc-tools/src/server/tonic/interceptor.rs
lvboudre, line 1: file to review here.

yellowstone-grpc-tools/src/server/window.rs
lvboudre, line 1: file to review.

yellowstone-grpc-geyser/src/config.rs
lvboudre, line 1: to review

yellowstone-grpc-e2e-test/README.md
lvboudre, line 1: to read.

yellowstone-grpc-geyser/src/ratelimit.rs
```diff
@@ -1,12 +1,494 @@
...
+                            self.window,
+                        ))
+                    })
+                    .ok_or(())?;
+
+                Ok::<_, ()>(MethodWindowState {
+                    limiter,
+                    last_seen: now,
+                })
+            });
```
Ozodimgba, line 216: the rate limit is resolved to build the key, do we need to repeat the same derivation logic to build the limiter...is there a reason we can't just grab the the value already sitting in `window_key.ratelimit`

yellowstone-grpc-geyser/src/ratelimit.rs
```diff
@@ -1,12 +1,494 @@
...
+                .as_ref()
+                .and_then(|ratelimits| {
+                    ratelimits
+                        .methods
+                        .get(request.uri().path())
+                        .or(Some(&ratelimits.default))
+                })
+                .copied()
+                .unwrap_or(self.max_hits_fallback),
+        };
```
Ozodimgba, line 186: something like

```rust
let resolved = subscription_info
    .ratelimits
    .as_ref()
    .and_then(|rl| rl.methods.get(path).or(Some(&rl.default)))
    .copied()
    .unwrap_or(self.max_hits_fallback);

let window_key = MethodRatelimitKey {
    ratelimit: resolved,
};

let mut state = self.shared.active_windows_map
    .entry(window_key)
    .or_insert_with(|| MethodWindowState {
        limiter: TokenBucketRateLimiter::new(resolved as u64, self.window),
        last_seen: now,
    });


```

yellowstone-grpc-geyser/src/ratelimit.rs
```diff
@@ -1,12 +1,494 @@
...
+                            .get(request.uri().path())
+                            .or(Some(&ratelimits.default))
+                    })
+                    .and_then(|ratelimit| {
+                        if *ratelimit < 0 {
+                            None
+                        } else {
+                            let casted_ratelimit = *ratelimit as u64;
+                            Some(TokenBucketRateLimiter::new(casted_ratelimit, self.window))
+                        }
```
Ozodimgba: we are repeating this guard no?

yellowstone-grpc-geyser/src/ratelimit.rs
```diff
@@ -1,12 +1,494 @@
...
+            });
+
+        let Ok(mut state) = result else {
+            log::info!(
+                "No rate limit for subscriber {} and method {}, skipping rate limiting logic",
+                subscription_info.subscription_id,
+                request.uri().path(),
+            );
+            return Ok(request);
+        };
```
Ozodimgba: based on the control flow i don think we can ever hit this error we are handling.

the window_key.ratelimit less than zero guard is above.

if self.max_hits_fallback < 0 is unreachable because if fallbacks were negative, window_key.ratelimit would
have been negative and caught by the guard

yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1604,6 +1605,14 @@ impl GrpcService {
...
         }
 
+        let method_ratelimit_config = auth.as_ref().and_then(|auth_config| match auth_config {
+            AuthConfig::Http(http_auth_config) => http_auth_config.ratelimit.clone(),
+            AuthConfig::File(file_auth_config) => file_auth_config.ratelimit.clone(),
+            AuthConfig::TrustedMetadata(trusted_metadata_config) => {
+                trusted_metadata_config.ratelimit.clone()
+            }
+        });
+
```
Ozodimgba: if all 3 structs carry `ratelimit` with the same type, it should be a method on `AuthConfig`

yellowstone-grpc-tools/src/server/tonic/interceptor.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn build(&self) -> Self::Interceptor;
+}
+
+impl<F, I> HttpInterceptorFactory for F
+where
+    F: Fn() -> I,
+    I: HttpInterceptor,
+{
+    type Interceptor = I;
+
```
Ozodimgba: nothing in the codebase so far calls `build()`, the server wires, MethodRatelimiter directly
via HttpInterceptorLayer::new(ratelimiter) and layer() clones the interceptor so it doesnt call it either.

yellowstone-grpc-geyser/src/config.rs
```diff
@@ -430,11 +430,65 @@ pub struct HttpBackedAuthConfig {
...
+        Duration::from_secs(10)
+    }
 }
 
 #[derive(Debug, Clone, Deserialize)]
 pub struct FileBackedAuthConfig {
     pub subscription_resolver_path: PathBuf,
+
+    #[serde(default)]
+    pub ratelimit: Option<RatelimitConfig>,
```
Ozodimgba: nit but I would add the  Option<RatelimitConfig> to AuthConfig itself or in a shared base struct at this point, the duplication also is what forces the 3 branch match in the wiring code too

yellowstone-grpc-geyser/src/config.rs
```diff
@@ -430,11 +430,65 @@ pub struct HttpBackedAuthConfig {
...
+            default_max_hits: 1000,
+            window: Duration::from_secs(10),
+        }
+    }
+}
+
+impl RatelimitConfig {
+    pub const fn default_max_hits() -> i32 {
+        1000
+    }
```
Ozodimgba, line 476: default stated twice. Self::default_max_hits() instead of hardcoding

yellowstone-grpc-geyser/src/config.rs
Ozodimgba: see https://github.com/rpcpool/yellowstone-grpc/pull/803#discussion_r3538233144

yellowstone-grpc-geyser/src/ratelimit.rs
lvboudre: You are right. It's bad code that I entangled myself in due to multiple iteration, Ill apply your proposal down below.

yellowstone-grpc-tools/src/server/tonic/interceptor.rs
lvboudre: Good catch, I forgot to remove this code, it comes from previous attempt.

yellowstone-grpc-geyser/src/config.rs
lvboudre: It's kinda because of serde that cannot use `Default` implementation.
But I will try to use `default_max_hits` inside the `default` block.

yellowstone-grpc-geyser/src/config.rs
lvboudre: done

Ozodimgba (approved):

PR #810, opened 2026-07-05
fix: block reconstruction delivery path correctness

Ozodimgba:
- fix(ReplayIter): use continue instead of ? on missing progression entry to prevent one slot's missing data from terminating the entire iterator
- fix(ReplayedSlot): extract into_messages() with correct ordering matching live broadcast path, parent_slot preservation, and Block/BlockMeta inclusion
- fix(ForksDetected/DeadSlotDetected): retain ready_queue entries so forked or dead blocks already queued do not escape to subscribers
- fix(prune_slot): recompute min_slot cache when pruned slot was the advertised minimum, preventing stale replay slot advertisement
- fix(handle_block_data): guard against creating phantom ProcessingSlot entries for already-frozen slots
- fix(seal): add symmetric entry count mismatch check

### DIFF
--- a/yellowstone-grpc-geyser/src/block_reconstruction.rs
+++ b/yellowstone-grpc-geyser/src/block_reconstruction.rs
@@ -213,24 +213,24 @@ pub struct ReplayedSlot<'frozen_block> {
 }
 
 impl<'frozen_block> ReplayedSlot<'frozen_block> {
-    /// Converts a replayed slot into the message sequence a subscriber expects.
-    pub fn to_messages(&self) -> Vec<Message> {
+    /// Extends the caller's buffer with the message sequence a subscriber expects.
+    /// Ordering matches the live broadcast path in block_reconstruction_loop.
+    pub fn extend_messages(&self, out: &mut Vec<Message>) {
         // Invariant: slot status messages carry their original parent_slot.
         // Invariant: content arrives before commitment signal.
-        let mut messages = Vec::new();
 
         // Block content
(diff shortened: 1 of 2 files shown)


yellowstone-grpc-geyser/src/block_reconstruction.rs
```diff
@@ -212,13 +212,52 @@ pub struct ReplayedSlot<'frozen_block> {
     pub slot_status_messages: Vec<SlotCommitmentStatusUpdate>,
 }
 
+impl<'frozen_block> ReplayedSlot<'frozen_block> {
+    /// Converts a replayed slot into the message sequence a subscriber expects.
+    pub fn to_messages(&self) -> Vec<Message> {
+        // Invariant: slot status messages carry their original parent_slot.
+        // Invariant: content arrives before commitment signal.
+        let mut messages = Vec::new();
```
Synestraa: This messages vec is being given to replay's Vec as part of an extend, the replay's vec has a capacity of 32768 which is not properly utilized

You should pass &mut Vec<Message> here and remove the return

Synestraa (changes requested):

yellowstone-grpc-geyser/src/block_reconstruction.rs
Ozodimgba: yeah there shouldn't be an intermediate allocation. thanks

Synestraa (approved):

PR #823, opened 2026-07-14
e2e: deshred ALT resolution scenario

leafaar:
## Summary

Adds a single e2e scenario that verifies address lookup table (ALT) support in the deshred stream, covering both resolution and ALT-aware filtering in one pass. Follow-up to the ALT resolution fix (#821).

## `deshred-alt-resolution`

Subscribes to deshred transactions filtered on common token mints (USDC, USDT, wSOL) via `account_include`, and in one scan verifies:

- **Resolution**: transactions referencing address table lookups come back with `loaded_writable_addresses` / `loaded_readonly_addresses` matching the lookup index counts. Deshred is pre-commitment and ALTs resolve against the rooted bank, and resolution is all-or-nothing per transaction, so the scenario requires a minimum number of fully resolved transactions rather than failing on the first miss. When resolution is disabled every such transaction comes back empty and the minimum is never reached.
- **ALT-aware filtering** (per review feedback): `account_include` matches a transaction where a mint is present only in the ALT-loaded addresses and not the static keys, proving the filter considers ALT-resolved accounts. Token mints are used because static accounts (System Program, Token program) never appear in an ALT and could not prove ALT matching; a single ALT-only match is sufficient.

Verified against a live endpoint running the fix: passes (repeatably). On an endpoint without the fix the resolution check fails (`resolved ALT addresses on only 0 of N ...`).

## Also included

- `geyser`: replace `message_batch.len() == 0` with `is_empty()` in `geyser_loop` (clears a clippy `len_zero` lint so the crate passes `clippy -D warnings`).

fmt and clippy clean on the touched crates.

### DIFF
--- a/yellowstone-grpc-e2e-test/src/scenarios/default.rs
+++ b/yellowstone-grpc-e2e-test/src/scenarios/default.rs
@@ -7,6 +7,7 @@ use {
         collections::{HashMap, HashSet},
         str::FromStr,
     },
+    tokio::time::{timeout, Duration},
     tokio_stream::StreamExt,
     yellowstone_grpc_e2e_macros::test_helper,
     yellowstone_grpc_geyser::plugin::message::CommitmentLevel,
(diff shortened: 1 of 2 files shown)


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1171,7 +1171,7 @@ impl GrpcService {
                 break;
             };
 
-            if message_batch.len() == 0 {
+            if message_batch.is_empty() {
```
leafaar, line 1174: clippy was reporting this so i fixed since its a small thing but it was left from master.

yellowstone-grpc-e2e-test/src/bin/yellowstone-e2e.rs
```diff
@@ -62,6 +62,11 @@ struct Cli {
     /// TOML config file for scenario-specific parameters.
     #[arg(long, value_name = "PATH")]
     config_file: Option<PathBuf>,
+
+    /// Before running scenarios, call the GetVersion unary RPC on the endpoint and print
+    /// the plugin version. Useful to confirm which build an endpoint is serving.
+    #[arg(long = "with-version")]
+    with_version: bool,
```
lvboudre: we don't need this. please remove it.

yellowstone-grpc-e2e-test/src/scenarios/default.rs
```diff
@@ -442,6 +442,123 @@ pub async fn test_subscribe_deshred(config: &RunConfig) -> Result<()> {
     Ok(())
 }
 
+/// Verifies address lookup table (ALT) resolution is enabled for deshred transactions:
+/// transactions referencing address table lookups come back with their writable/readonly
+/// loaded addresses resolved, matching the lookup index counts. When resolution is disabled
+/// those addresses stay empty, so no transaction resolves and this scenario fails.
+#[test_helper(name = "deshred-alt-resolution", tags = ["deshred", "alt"])]
+pub async fn deshred_should_resolve_address_lookup_tables(config: &RunConfig) -> Result<()> {
```
lvboudre: We should also test that filters over account include must match account exclusively present in ALT.

lvboudre (changes requested):

lvboudre (approved):

PR #832, opened 2026-07-22
Message delivery speed in processed

Synestraa:
Moved Arc<> to outer objects in Message enum type, reducing overall size of each send() and recv()


lvboudre (commented):
It must 14.2.2 not 14.2.1

lvboudre (approved):

PR #839, opened 2026-07-28
Fix dedup dropping messages after BlockMeta

senzenn:
Follow-up to #717 and #753. DedupStream wraps every subscription, reconnect enabled or not. Four pre-existing bugs:

1. Messages arriving after their slot's BlockMeta were quarantined pending a second BlockMeta that only a replay produces, so in steady state they were dropped. This swallowed the freeze-time account writes #838 fixed server-side.
2. DedupState.sealed grew without bound: a slot whose first message is BlockMeta never entered slot_order, which is all prune() evicts from.
3. DedupKey was positional for transactions, transaction statuses and entries, so two blocks for one slot collided. Now content-addressed.
4. Slots in flight at reconnect lost their delivered keys, so replayed content was flushed wholesale and re-delivered.

Testing

New e2e drives the client against an in-process Geyser server through forced disconnects. Against master: 40 lost messages plus duplicates. Sweeps 55 disconnect offsets, passes at each.

Review notes

- #753 owner please confirm. Partial slots now dedup on delivered keys instead of flushing wholesale.
- Account dedup assumes write_version is stable across a replay. True here (grpc.rs:1403); a server that regenerates messages would emit duplicates.
- Two existing tests asserted the old partial-slot behaviour and were replaced

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,20 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+- yellowstone-grpc-client 13.2.2
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


senzenn:
@linuskendall  @lvboudre

lvboudre:
Fixed

PR #842, opened 2026-07-29
STR-620: refactor(client) - decouple reconnect from gap recovery

Ozodimgba:
Today users who opt out of auto-reconnect are still coupled to it internally, so they pay for dedup they never use. This splits GeyserStream into variants, one of which is the bare gRPC stream, so opting out actually opts you out. It also
lifts DedupStream out of AutoReconnect, which deletes the state handoff between them and gives us a third variant almost for free: reconnect without dedup, for consumers that want to auto-reconnect but have no use for the messages lost in disconnection gap.

Extends and Closes #STR-620

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ### Features
 
-- Added `ReconnectionPolicy::NoReplay` so a subscription can auto-reconnect from the tip without paying for replay or dedup.
+- Added `ReconnectionPolicy` so a subscription can choose whether or not to auto-reconnect from the tip without paying for replay or dedup.
 
 ### Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 - yellowstone-grpc-client 13.3.0
 - yellowstone-grpc-client-simple-12.3.1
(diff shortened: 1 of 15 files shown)


leafaar (approved):

examples/rust/src/bin/client_with_reconnect.rs
```diff
@@ -28,16 +36,30 @@ struct Args {
...
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
     env_logger::init();
     let args = Args::parse();
 
+    let reconnect_config = ReconnectConfig {
+        backoff: Backoff::default(),
+        policy: match args.policy {
+            Policy::Recover => ReconnectionPolicy::RecoverMissedData {
+                slot_retention: 250,
```
WilfredAlmeida: Why hardcoded 250? Shouldn't this come from the args as well?

Ozodimgba: Well hmmm thats true, i should export the default from reconnect.

CHANGELOG.md
```diff
@@ -10,6 +10,22 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+- yellowstone-grpc-client 13.3.0
+- yellowstone-grpc-client-simple-12.3.1
+
```
WilfredAlmeida, line 16: Also add for yellowstone-grpc-client-nodejs

WilfredAlmeida (approved):

PR #843, opened 2026-08-03
Separating finalized/confirmed/processed broadcast

Synestraa:


yellowstone-grpc-geyser/src/grpc.rs
```diff
@@ -1505,6 +1555,12 @@ impl GrpcService {
                             session.set_filter(filter_new);
                             info!("client #{}: filter updated", session.subscriber_id);
 
+                            let commitment_new = session.filter.get_commitment_level();
+                            if commitment_new != commitment {
+                                commitment = commitment_new;
+                                messages_rx = broadcast.subscribe(commitment);
+                            }
+
```
lvboudre, line 1565: When you subscribe to a new commitment, you don't drain any message you may have received between the last recv and the new subscribe. So user can actually lose data. 
Maybe this is fine, as I never saw anyone changing commitment inflight.

lvboudre (approved):

leafaar (approved):

PR #847, opened 2026-08-07
Str 638: Filter path optimizations: transaction include matching, empty-filter guard

Ozodimgba:
Closes #STR-638

### DIFF
--- a/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
+++ b/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
@@ -1597,20 +1597,19 @@ impl FilterBlocksInner {
         false
     }
 
-    fn matches_any_in_set(&self, account_keys: &FoldHashSet<Pubkey>) -> bool {
+   fn matches_any_in_set(&self, account_keys: &FoldHashSet<Pubkey>) -> bool {
         if self.account_include.is_empty() && self.account_cuckoo.is_none() {
             return true;
         }
 
-        if !self.account_include.is_empty() {
-            let (small, large) = if self.account_include.len() <= account_keys.len() {
-                (&self.account_include, account_keys)
-            } else {
-                (account_keys, &self.account_include)
-            };
-            if small.iter().any(|k| large.contains(k)) {
-                return true;
-            }
+        if !self.account_include.is_empty()
+            && self
+                .account_include
+                .intersection(account_keys)
+                .next()
+                .is_some()
+        {
+            return true;
         }
 
         if let Some(cuckoo) = &self.account_cuckoo {
--- a/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
+++ b/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
@@ -46,6 +46,45 @@ use {
     },
 };
 
+/// Solana caps a transaction at 64 account keys, so a filter list at or
+/// below this is never longer than the transaction it is compared against.
+const MAX_TX_ACCOUNT_KEYS: usize = 64;
+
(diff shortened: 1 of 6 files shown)


yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -46,6 +46,45 @@ use {
...
+/// below this is never longer than the transaction it is compared against.
+const MAX_TX_ACCOUNT_KEYS: usize = 64;
+
+#[derive(Debug, Clone)]
+enum HybridSet<T> {
+    /// Cheap to walk.
+    Contiguous(Vec<T>),
+    /// Cheap to probe.
+    Hashed(FoldHashSet<T>),
+}
```
lvboudre, line 59: Please, just use FoldHashSet, seems over-engineered solution,
a FoldHashSet should be backed by a vec that is probably filled up to 75% of the capacity.
so you could just with_capacity(64 + 25% buffer).

If entirely fileld with 64 elemetns, traversing should be fast too.

Also, what gains did you get from doing that is worth changing 150LOC ?

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -1058,18 +1104,39 @@ impl FilterTransactions {
...
 
-                if !inner.account_include.is_empty()
-                    && !inner.account_include.iter().any(in_effective_set)
-                {
-                    return None;
+              if !inner.account_include.is_empty() {
+                    let hit = match &inner.account_include {
+                        HybridSet::Contiguous(v) => v.iter().any(in_effective_set),
+                        HybridSet::Hashed(s) => effective_keys().any(|k| s.contains(k)),
+                    };
```
lvboudre: If you keep the "HybridSet", this should be put inside a method instead of sprawling code about variants at every callsite.

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -1058,18 +1104,39 @@ impl FilterTransactions {
...
+                        return None;
+                    }
                 }
 
-                if inner.account_exclude.iter().any(in_effective_set) {
-                    return None;
+                if !inner.account_exclude.is_empty() {
+                    let hit = match &inner.account_exclude {
+                        HybridSet::Contiguous(v) => v.iter().any(in_effective_set),
+                        HybridSet::Hashed(s) => effective_keys().any(|k| s.contains(k)),
```
lvboudre: the match expression should be inside a method of `HybridSet`

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
Ozodimgba: honestly, i was really calm with foldhash but I dont think this is "over-engineered". this enum maybe shouldn't be here...although I think srdja was right about sorting out the regressesion.

so it really is about "adding" the 150 LOC rather than "changing"...we didnt change that many lines

Ozodimgba: correct me if i am wriong but I was under the impression that hashbrown rounds up to a power of two, so with_capacity(80) allocates 128 slots and 64 elements sits at 50% occupancy, not 75. iteration walks control bytes for all 128 either way. happy to bench it if you want the number

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
Ozodimgba: thanks

Synestraa (approved):

PR #850, opened 2026-08-11
STR-636: Added client test to e2e

Ozodimgba:


yellowstone-grpc-e2e-test/src/scenarios/default/reconnect.rs
```diff
@@ -0,0 +1,217 @@
...
+                filter_by_commitment: Some(true),
+                interslot_updates: Some(false),
+            },
+        )]),
+        blocks_meta: HashMap::from([(
+            "test".to_string(),
+            SubscribeRequestFilterBlocksMeta::default(),
+        )]),
+        commitment: Some(CommitmentLevel::Processed as i32),
+        ..Default::default()
```
lvboudre, line 76: I would like to see tests that also subscribe to transactions/accounts please.
We should also adds test for other commitment level like Confirmed at least too.

lvboudre (changes requested):
I would like to see a test with blockmachine too.
See if you can build blocks with exepcted transactions count and entries count with an unstable grpc connetor too please.

yellowstone-grpc-e2e-test/src/scenarios/default/reconnect.rs
```diff
@@ -0,0 +1,375 @@
...
+                            matched += 1;
+                        }
+                    }
+                }
+                block = blocks.next() => {
+                    let output = block.context("block stream ended")??;
+                    if let BlockMachineOutput::FrozenBlock(b) = output {
+                        let slot = b.slot;
+                        log::info!("rebuilt block for slot {slot}");
+                        built.insert(slot, b);
```
lvboudre, line 341: you should assert the block was never insert twice

lvboudre (dismissed):

leafaar (approved):

leafaar (approved):

PR #856, opened 2026-08-17
Account filtering performance

Synestraa:
Introduced performance increases on account filtering, tested thoroughly for both compatibility and performance across hundreds of different variations of account filter types.

In the end, benched performance improvement is anywhere between 1.5x to 723x (large amount of filters are negated almost entirely).

### DIFF
--- a/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
+++ b/yellowstone-grpc-geyser/src/plugin/filter/filter.rs
@@ -14,8 +14,9 @@ use {
             name::{FilterName, FilterNameError, FilterNames},
         },
         message::{
-            CommitmentLevel, Message, MessageAccount, MessageBlock, MessageBlockMeta,
-            MessageDeshredTransaction, MessageEntry, MessageSlot, MessageTransaction, SlotStatus,
+            CommitmentLevel, Message, MessageAccount, MessageAccountInfo, MessageBlock,
+            MessageBlockMeta, MessageDeshredTransaction, MessageEntry, MessageSlot,
+            MessageTransaction, SlotStatus,
         },
     },
     base64::{engine::general_purpose::STANDARD as base64_engine, Engine},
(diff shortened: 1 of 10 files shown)


lvboudre:
@claude review

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -463,70 +442,332 @@ impl Filter {
...
+
+impl FilterBitSet {
+    const BITS_PER_WORD: usize = u64::BITS as usize;
+
+    fn new(bits: usize) -> Self {
+        Self(vec![0; bits.div_ceil(Self::BITS_PER_WORD)].into_boxed_slice())
+    }
+
+    fn insert(&mut self, index: usize) {
+        self.0[index / Self::BITS_PER_WORD] |= 1 << (index % Self::BITS_PER_WORD);
```
lvboudre: This function has no boundary check, it would wise to add asserts to make sure callsite does not try to access out-of-bound position.

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -463,70 +442,332 @@ impl Filter {
...
 }
 
+#[derive(Debug, Clone)]
+struct FilterBitSet(Box<[u64]>);
+
+impl FilterBitSet {
+    const BITS_PER_WORD: usize = u64::BITS as usize;
+
+    fn new(bits: usize) -> Self {
+        Self(vec![0; bits.div_ceil(Self::BITS_PER_WORD)].into_boxed_slice())
```
lvboudre, line 556: I would assert `bits` is a multiple of `u64::BITS`

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
```diff
@@ -463,70 +442,332 @@ impl Filter {
...
+    owner_fallback: FilterBitSet,
+
+    account_fallback_empty: bool,
+    owner_fallback_empty: bool,
+}
+
+impl FilterAccountsIndex {
+    const MAX_BITSET_BYTES: usize = 512 * 1024 * 1024;
+
+    fn new(aggregates: &[FilterAccountAggregate]) -> Option<Self> {
```
lvboudre: It should not be the responsability of the constructor to decide if yes or not it can construct itself.
It's quite unusual to have Option<Self> as the output of `new`. This check should be done by the caller.
The caller must decide if it wants to build an AccountsIndex or not. Otherwise, it spread to construction logic at two places.

lvboudre: You could rename `new` with `try_init(...) -> Result<Self, AccountsIndexTryInitError> too.

yellowstone-grpc-geyser/src/plugin/filter/filter.rs
Synestraa: Bits does not need to be a multiple of u64::bits or any particular value

leafaar (approved):
lgtm!

lvboudre (approved):

Issue #761, opened 2026-06-11
Node client: native (off-heap) memory leak persists in 5.0.9 — reduced ~57% by #755, not eliminated


LamboPoewert:
### Summary
The NAPI Node client (`@triton-one/yellowstone-grpc`) leaks **native, off-heap** memory under a sustained high-throughput subscription: process RSS climbs roughly linearly while the V8 heap stays flat. **5.0.9** (#755, which moved decoding to JS-side protobuf buffers) **reduced the rate ~57% but did not eliminate it.**

This is distinct from #582 / #567 (validator / geyser-plugin RAM) — it's the **Node client process**.

### Environment
- `@triton-one/yellowstone-grpc` **5.0.9** (engine `yellowstone-grpc-napi-linux-x64-gnu@0.0.14`)
- Node **v22.22.2**, Linux x64 (glibc)
- One `subscribe()` stream, `CONFIRMED` commitment, high-throughput transaction filter

### Measurements
| version | RSS growth | heapUsed |
|---|---|---|
| 5.0.8 | ~1.4 GB/h | flat |
| 5.0.9 | **~0.6 GB/h** (linear, no plateau over a 2h window) | flat |

RSS climbs steadily; `heapUsed` and GC stay stable → the growth is off-heap (native / glibc), not V8.

### Why #755 looks relevant
#755 removed the Rust→JS object-construction path (`js_types.rs` / `typegen.rs`) in favor of returning encoded protobuf buffers decoded on the JS side. The leak **halving** after that change suggests a meaningful portion lived in that cross-NAPI object construction — but a **residual ~0.6 GB/h remains outside what #755 touched**.

### Reproduction
```js
import Client, { CommitmentLevel } from "@triton-one/yellowstone-grpc";

const c = new Client(process.env.ENDPOINT, process.env.TOKEN, { "grpc.max_receive_message_length": 64 * 1024 * 1024 });
await c.connect();
const stream = await c.subscribe();

// Full transaction firehose (no account filter) for max throughput.
stream.write({
  accounts: {}, slots: {}, blocks: {}, blocksMeta: {}, entry: {}, transactionsStatus: {}, accountsDataSlice: [],
  transactions: { all: { accountInclude: [], accountExclude: [], accountRequired: [], vote: false, failed: false } },
  commitment: CommitmentLevel.CONFIRMED,
}, (e) => e && console.error("write error:", e.message));

let txs = 0;
[...]

PR #863, opened 2026-08-20
nodejs: bound the subscribe worker→JS channel so slow consumers backpressure the server

### LINKED ISSUE #854: Node client: no backpressure — `DuplexStream` buffers updates in an unbounded native channel; proposal for bounded channel + batched `read()`
### Summary

The NAPI Node client has no end-to-end backpressure. In `yellowstone-grpc-client-nodejs/napi/src/lib.rs`, `DuplexStream::subscribe` spawns a worker that drains the tonic stream as fast as the wire delivers and pushes each encoded update into an **unbounded** tokio mpsc channel (the existing `// TODO : Fine tune unbounded channels.`, currently `lib.rs:177-178` on master):

```rust
// TODO : Fine tune unbounded channels.
let (readable_tx, readable_rx) = unbounded_channel::<Vec<u8>>();
```

The JS side consumes via `read()` one message per NAPI hop. Whenever the JS consumer is even slightly slower than the subscription's wire rate, the deficit accumulates in this native channel indefinitely:

- Process RSS grows without bound while the V8 heap stays flat (native `Vec<u8>` backlog, invisible to JS heap tooling).
- tonic/h2 flow control never engages — the worker's `stream_rx.next()` loop always polls ready, so the HTTP/2 receive window never closes and the server keeps sending at full rate.
- `stream.pause()` on the JS `Duplex` wrapper is a no-op for the native side: it only stops JS-side `read()` pulls, the Rust worker keeps buffering.
- The only remediation is `destroy()` + resubscribe, which discards the entire buffered backlog (data loss for the consumer).

This is likely a contributor to the "linear RSS growth, flat heapUsed" reports in #761 — any sustained consumer deficit on a high-throughput subscription manifests exactly like a native leak.

### Environment / measurements

- `@triton-one/yellowstone-grpc` 5.0.9 (`yellowstone-grpc-napi-linux-x64-gnu`), Node 22, Linux x64.
- Sustained high-throughput transaction subscription (post-filter wire rate ~1.8 MB/s); a consumer sustaining ~6–7% of the wire rate accumulates multi-GB native backlog between process restarts; RSS floor ratchets up ~500 MB per pause/resume cycle and never returns.
- Confirmed the backlog lives on the native side: heap snapshots flat, `rss - heapTotal - external` grows linearly with subscription uptime.

### Proposal

Two complementary changes, both backward-compatible:

1. **Bounded readable channel with configurable capacity.** Replace `unbounded_channel::<Vec<u8>>()` with `mpsc::channel(capacity)` and `readable_tx.send(update_bytes).await` in the worker loop. When JS is slow, the worker suspends between `stream_rx.next()` polls, tonic stops reading from the h2 stream, the HTTP/2 window fills, and the server throttles — real end-to-end backpressure with zero API change. Capacity could default high (e.g. 4096 messages) to preserve current behavior in practice, exposed as a `subscribe()` option (e.g. `{ readableCapacity?: number }`).

   Note: consumers that fall behind *permanently* will then lag on the server side instead (and eventually hit the server's own lagged-slot handling), but that is the correct failure mode — visible, bounded, and recoverable — versus unbounded client memory growth.

2. **Batched handoff: `read_many(maxItems, maxBytes)`.** A single NAPI crossing returning up to N encoded payloads (e.g. as an array of Buffers or one concatenated length-prefixed Buffer). At firehose rates the per-message NAPI hop + Promise round-trip is measurable overhead; batching amortizes it and materially raises the max sustainable JS-side consumption rate. `read()` can remain as `read_many(1)`.

Happy to contribute a PR for either or both if the direction is acceptable — (1) is a small, contained change to `DuplexStream::subscribe` (and the deshred variant at `lib.rs:424-425`).

### Related

- #761 (native RSS growth on the Node client — a slow-consumer backlog in this channel reproduces those symptoms exactly)
- #435 added `channel_capacity_filters` on the geyser side; this is the client-side analogue.

kostia-lev:
Closes #854 (and, we believe, explains #761).

## Problem

The NAPI subscribe worker forwards every `SubscribeUpdate` into an
`unbounded_channel::<Vec<u8>>()` (`napi/src/lib.rs`, carrying the in-tree
`// TODO : Fine tune unbounded channels.`), and polls the tonic stream
unconditionally. Two consequences:

- h2 flow control never engages: a JS consumer slower than the stream cannot
  slow the server down, no matter what it does (`pause()`, slow `read()`s —
  nothing reaches the socket).
- The surplus accumulates as native memory, invisible to the V8 heap, until
  the process is OOM-killed. This is the "native off-heap memory leak"
  symptom profile reported in #761: RSS climbs linearly while `heapUsed`
  stays flat.

We run this client in production against a Solana mainnet firehose
(pump.fun/PumpSwap-wide filters, hundreds of updates/s, multi-GB `fromSlot`
replay bursts). Whenever downstream persistence stalled for minutes, the
channel absorbed the difference: 12–18 GB RSS with ~700 MB of V8 heap,
ending in OOM kills.

## Change

- The readable channel is bounded: `mpsc::channel(capacity)` instead of
  `unbounded_channel()`, for both `DuplexStream` and `DuplexStreamDeshred`.
- The worker reserves send capacity **before** polling the gRPC stream, and
  polls `stream_rx` only while a permit is held. Holding no permit leaves
  the socket unread, so tonic stops draining it and the h2 receive window
  closes — the server is throttled at the source. `writable_rx` stays in the
  same `select!`, so subscription updates written from JS are serviced even
  while fully backpressured.
- Capacity is a new channel option, `grpcSubscribeReadableChannelCapacity`
  (default **1024** updates; an explicit **0** restores the previous
  unbounded behavior), resolved once per client following the same
  default-then-override convention as the other options.

No JS API changes beyond the new optional field; the generated typings are
included.

## Measurements (production endpoint, mainnet firehose)

Full production subscription (two tx filters incl. a 30k-address
`accountInclude`, two account subs with memcmp filters) + 750-slot
`fromSlot` replay, pulled via `read()`:

| consumer | unpatched | patched (capacity 1024) |
|---|---|---|
| full speed | RSS flat ~238 MB, 118k frames/30 s | RSS flat ~175 MB, 120k frames/30 s |
| deliberately slow (50 reads/s vs the dump) | native backlog grows unboundedly | **RSS flat ~110 MB**; server throttled via h2; no disconnect, no data loss, stream resumes full rate when reads resume |

The slow-consumer case is the one that matters: today it is an OOM; with
this change it is ordinary gRPC flow control. We measured no throughput
penalty at full speed, and Helius (LaserStream) served a fully backpressured
subscription for extended periods without evicting the client.

## Notes for reviewers

- The default changes behavior (bounded 1024 vs unbounded). We think that is
  the right default — the unbounded channel is an OOM in disguise — but if
  you prefer strict compatibility, flipping the `JsChannelOptions::default()`
  value to `Some(0)` preserves today's behavior and makes the bound opt-in;
  the rest of the change is identical. Happy to amend.
- `1 << 20` as the "effectively unbounded" capacity keeps the channel type
  uniform instead of branching on bounded/unbounded variants.
- Existing test suite passes unchanged apart from the fixtures constructing
  the readable side directly; four new tests cover capacity resolution.
- Related observation from the same production migration, for anyone
  consuming the private NAPI surface: `subscribe_raw` now eagerly sends
  `SubscribeRequest::default()` when subscribed without an initial request,
  so a request written *after* open arrives as a filter update — and at
  least one provider treats `from_slot` on an update as short-range gap
  recovery rather than archive replay, killing the stream. Passing the
  initial request into `subscribe()` avoids this; may be worth a README
  note, though it is out of scope for this PR.

### DIFF
--- a/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
+++ b/yellowstone-grpc-client-nodejs/napi/src/bindings.rs
@@ -38,6 +38,16 @@ pub struct JsChannelOptions {
   pub grpc_max_decoding_message_size: Option<u32>,
   pub grpc_max_encoding_message_size: Option<u32>,
   pub grpc_default_compression_algorithm: Option<JsCompressionAlgorithm>,
+  /// Capacity, in updates, of the per-subscription worker -> JS channel.
+  ///
+  /// Bounded so that a JS consumer slower than the stream backpressures the
+  /// server: once the channel is full the worker stops polling the gRPC
+  /// stream, tonic stops reading the socket and the h2 receive window
+  /// closes. An unbounded channel instead accumulates the surplus in native
+  /// memory, invisible to the V8 heap, until the process is killed.
+  ///
+  /// `0` selects an effectively unbounded capacity (the previous behavior).
+  pub grpc_subscribe_readable_channel_capacity: Option<u32>,
   //--------------------
   // Flags.
   //--------------------
@@ -86,6 +96,7 @@ impl Default for JsChannelOptions {
   ///   grpc_initial_stream_window_size: Some(4 * 1024 * 1024),     // 4MB
   ///   grpc_max_decoding_message_size: Some(1 * 1024 * 1024 * 1024), // 1GB
   ///   grpc_max_encoding_message_size: Some(32 * 1024 * 1024),     // 32MB
+  ///   grpc_subscribe_readable_channel_capacity: Some(1024),        // updates
   ///   grpc_http2_adaptive_window: Some(true),
   ///   grpc_tcp_nodelay: Some(true),
   ///   grpc_keep_alive_while_idle: None,
@@ -105,6 +116,7 @@ impl Default for JsChannelOptions {
       grpc_initial_stream_window_size: Some(4 * 1024 * 1024),     // 4MB
       grpc_max_decoding_message_size: Some(1024 * 1024 * 1024),   // 1GB
       grpc_max_encoding_message_size: Some(32 * 1024 * 1024),     // 32MB
+      grpc_subscribe_readable_channel_capacity: Some(1024),       // updates
       grpc_http2_adaptive_window: Some(true),
       grpc_tcp_nodelay: Some(true),
       grpc_keep_alive_while_idle: None,
--- a/yellowstone-grpc-client-nodejs/napi/src/client.rs
+++ b/yellowstone-grpc-client-nodejs/napi/src/client.rs
@@ -46,6 +46,30 @@ where
   })
 }
 
+/// Resolves `grpc_subscribe_readable_channel_capacity` following the same
+/// convention as the other channel options: the `JsChannelOptions::default()`
+/// value applies unless the caller set the field explicitly. An explicit `0`
+/// selects a capacity so large it is unbounded in practice, restoring the
+/// previous behavior without a separate flag.
+fn resolve_subscribe_readable_channel_capacity(
+  channel_options: Option<&JsChannelOptions>,
+) -> usize {
+  /// One update is tens of KB, so this is tens of GB of buffering: never
+  /// reachable in practice, which is what makes it a faithful stand-in for
+  /// "unbounded" while keeping the channel type uniform.
+  const EFFECTIVELY_UNBOUNDED: usize = 1 << 20;
+
+  let capacity = channel_options
+    .and_then(|options| options.grpc_subscribe_readable_channel_capacity)
+    .or(JsChannelOptions::default().grpc_subscribe_readable_channel_capacity)
+    .unwrap_or(1024);
+  if capacity == 0 {
+    EFFECTIVELY_UNBOUNDED
+  } else {
+    capacity as usize
+  }
+}
+
 /// Main client struct exposed to JavaScript via NAPI.
 ///
 /// The client maintains a persistent gRPC connection that is created once
@@ -53,6 +77,9 @@ where
 #[napi]
 pub struct GrpcClient {
   pub(crate) client: GeyserGrpcClient,
+  /// Resolved `grpc_subscribe_readable_channel_capacity`; consumed by
+  /// `DuplexStream::subscribe` / `DuplexStreamDeshred::subscribe`.
+  pub(crate) subscribe_readable_channel_capacity: usize,
 }
 
 #[napi]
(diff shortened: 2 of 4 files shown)


lvboudre:
Can you sign your commit please.

yellowstone-grpc-client-nodejs/napi/src/bindings.rs
```diff
@@ -105,6 +116,7 @@ impl Default for JsChannelOptions {
       grpc_initial_stream_window_size: Some(4 * 1024 * 1024),     // 4MB
       grpc_max_decoding_message_size: Some(1024 * 1024 * 1024),   // 1GB
       grpc_max_encoding_message_size: Some(32 * 1024 * 1024),     // 32MB
+      grpc_subscribe_readable_channel_capacity: Some(1024),       // updates
```
lvboudre: Don't make it an Option, let's declare a default constant of 1024 instead.

yellowstone-grpc-client-nodejs/napi/src/client.rs
```diff
@@ -46,13 +46,40 @@ where
...
+    .and_then(|options| options.grpc_subscribe_readable_channel_capacity)
+    .or(JsChannelOptions::default().grpc_subscribe_readable_channel_capacity)
+    .unwrap_or(1024);
+  if capacity == 0 {
+    EFFECTIVELY_UNBOUNDED
+  } else {
+    capacity as usize
+  }
+}
+
```
lvboudre: Goes with comment above, we don't need an option for capacity, lets put smart default instead of 1024.

yellowstone-grpc-client-nodejs/napi/src/client.rs
```diff
@@ -75,6 +102,9 @@ impl GrpcClient {
   ) -> napi::Result<Self> {
     init_crypto_provider();
 
+    let subscribe_readable_channel_capacity =
+      resolve_subscribe_readable_channel_capacity(channel_options.as_ref());
+
```
lvboudre: if capacity is 0, panic.

lvboudre:
I am closing this issue since it will be fixed in #864  (removes unecessary intermediate unbounded channels).
This should rely on the actual GeyserStream backpressured gRPC stream.

PR #864, opened 2026-08-25
str-679: fix lack of backpressure in typescript client + refactor codes.

### LINKED ISSUE #854: Node client: no backpressure — `DuplexStream` buffers updates in an unbounded native channel; proposal for bounded channel + batched `read()`

lvboudre:
This PR fixes #854 by removing the intermediate unbound channels that used to exists in `DuplexStream` that bridges GrpcStream and Javascript world, where in fact, we could have just polled directly from the underlying `GeyserStream` object which already implements back pressure (backed by the underlying tonic's `Streaming` stream object).

This PR also removes duplicate code across `DuplexStream` and `DuplexStreamDeshred` using a common `DuplexStreamInner`.

Removes the usage of slow tokio's `Mutex` in favor of standard's lib `Mutex` object by building custom `Future` type which prevent cross-await point mutex locking and limits the lock lifetime to the `poll` scope only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,15 @@ The minor version will be incremented upon a breaking change and the patch versi
 ## 2026-08-25
 
 - yellowstone-grpc-client-nodejs 7.0.0
+- yellowstone-grpc-client 13.4.0
 
 ### Fixes
 
 - Fixes backpressure issue [#854](https://github.com/rpcpool/yellowstone-grpc/issues/854)
+
+### Features
+
+- new `Mock` variant for `GeyserStream` in `yellowstone-grpc-client` when `test-tools` feature-flag is enabled.
 
 ### Breaking
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+## 2026-08-25
+
+- yellowstone-grpc-client-nodejs 7.0.0
+
+### Fixes
+
+- Fixes backpressure issue [#854](https://github.com/rpcpool/yellowstone-grpc/issues/854)
+
+### Breaking
+
+- The `napi`'s `DuplexStream` and `DuplexStreamDeshred` `writes_raw` method now returns a `Promise` instead of being synchronous call.
+
 ## 2026-08-19
 
 - yellowstone-grpc-client-nodejs-6.0.0
--- a/yellowstone-grpc-client-nodejs/napi/src/lib.rs
+++ b/yellowstone-grpc-client-nodejs/napi/src/lib.rs
@@ -16,17 +16,17 @@ mod encoding;
 mod subscribe_request_validation;
 mod utils;
 
+use futures::{future::poll_fn, Sink, Stream, TryStream, TryStreamExt};
 use futures_util::{SinkExt, StreamExt};
 use napi::{bindgen_prelude::*, Env};
 use napi_derive::napi;
 use prost::Message;
-use std::sync::{
-  atomic::{AtomicBool, Ordering},
-  Arc, Mutex as StdMutex, Once,
+use std::{
+  sync::{Arc, Mutex as StdMutex, Once},
+  task::Poll,
 };
-use tokio::sync::{
-  mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
-  Mutex,
+use yellowstone_grpc_client::{
+  GeyserStream, SubscribeDeshredRequestSink, SubscribeDeshredStream, SubscribeRequestSink,
 };
 use yellowstone_grpc_proto::prelude::*;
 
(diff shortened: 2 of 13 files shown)


yellowstone-grpc-client-nodejs/napi/src/lib.rs
lvboudre, line 1: this is most of the change.

yellowstone-grpc-client-nodejs/napi/src/lib.rs
```diff
@@ -91,192 +91,271 @@ fn napi_error(status: napi::Status, reason: impl Into<String>) -> napi::Error {
...
+  St: Stream + Unpin,
+{
+  type Item = St::Item;
+
+  fn poll_next(
+    self: std::pin::Pin<&mut Self>,
+    cx: &mut std::task::Context<'_>,
+  ) -> std::task::Poll<Option<Self::Item>> {
+    let this = self.get_mut();
+    let mut guard = this.inner.lock().expect("state lock");
```
WilfredAlmeida: If this error is triggered, is it a failure the user should know about? If yes, will it bubble up?

CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ The minor version will be incremented upon a breaking change and the patch versi
 
 ## [Unreleased]
 
+## 2026-08-25
+
+- yellowstone-grpc-client-nodejs 7.0.0
+
```
WilfredAlmeida, line 17: Also needs entry about rust `yellowstone-grpc-client` bump to `v13.4.0`

WilfredAlmeida (approved):

Synestraa (approved):

leafaar (approved):

yellowstone-grpc-client-nodejs/napi/src/lib.rs
lvboudre: yes

Issue #867, opened 2026-08-26
Move the proto-to-Solana-type conversion out of the plugin crate so clients can use it


serejke:
### Describe your build and run environment

Source-level proposal against `master` (dc69138); not tied to a runtime environment or config. Rust 1.96.1 per `rust-toolchain.toml`.

### Describe your issue

**Motivation**

This project maintains its own protobuf schema and its own converters between that schema and Solana types. Both are the right call: the wire format serves four languages, and it evolves on a different clock from agave's storage schema, which is effectively frozen by the block archive it encodes.

The converters, though, live in the validator-plugin crate — so the clients this project exists to serve cannot reach them. The Node client borrows agave's converter instead. The net effect is that the project pays the full cost of owning a converter and still carries the risks of depending on someone else's.

Conversion between this repo's wire format and Solana types is a property of the protocol, not of the plugin. It should live where every consumer of the protocol can use it.

```
TODAY
                     +--------------------------------+
  wire --- proto --->| yellowstone-grpc-geyser        |--> Solana types --> Rust clients
                     |   convert_from   (correct)     |
                     +--------------------------------+
                          ^
                          X  a Node binding must not depend on the plugin crate
                          |
                     +--------------------------------+
  wire --- proto --->| napi                           |--> Solana types --> Node client
                     |   re-encode -> re-parse as     |
                     |   agave's twin -> agave's From |
                     +--------------------------------+

PROPOSED
                     +--------------------------------+
  wire --- proto --->| yellowstone-grpc-convert (new) |--> Solana types
                     |   convert_to / convert_from    |
                     +--------------------------------+
                          ^                    ^
                          |                    |
                  yellowstone-grpc-geyser     napi          one implementation,
                        (plugin)          (Node client)     reachable by every consumer
```

**What the detour costs today**

1. It inherits agave's conversion bugs. Agave's inbound conversion reads only the `versioned` flag and never `config`, so V1 transactions are reconstructed as V0 with the compute-budget request dropped (#866; upstream anza-xyz/agave#14873, fix anza-xyz/agave#14874). The bug this repo already fixed in #855 still reaches Node users by the other route.
2. It rests on this repo's `solana-storage.proto` and agave's staying wire-identical, which nothing enforces or tests. They have already drifted textually — `optional TransactionConfig config = 7`, a different declaration order, a `go_package` option — benign on the wire today, but nothing would catch a drift that is not.
[...]

Issue #866, opened 2026-08-26
Node client: txEncode.encode reports V1 transactions as version 0 with transactionConfig absent


serejke:
### Versions

* `@triton-one/yellowstone-grpc` 6.0.0 (Node client)
* Verified empirically with a real V1 transaction produced on a local `solana-test-validator` 4.2.1

### Describe your issue

`txEncode.encode` reports V1 transactions as `version: 0` with `transactionConfig` absent. No error is raised.

Cause: in `yellowstone-grpc-client-nodejs/napi/src/encoding.rs`, both `encode_tx` and `encode_deshred_tx` re-encode the yellowstone proto message and decode it as `solana-storage-proto`'s type. That crate's inbound conversion (`From<generated::Message> for VersionedMessage`) branches only on the `versioned` boolean and never reads `config`, so every V1 message is reconstructed as V0 and the config is dropped. This is the same `versioned`-only bug you already fixed in your own `convert_from` in #855 (released in `yellowstone-grpc-geyser` 15.1.1) — but the Node client path goes through `solana-storage-proto` (4.2.0 and current master), which still has it.

A fix is open upstream in agave: anza-xyz/agave#14873, PR anza-xyz/agave#14874.

This issue is informational: once the upstream fix lands and the dependency is bumped, the Node client path is corrected with no changes needed here.

serejke:
Fixed upstream, as @lvboudre preferred — thanks for raising it with Anza directly.

* agave [#14874](https://github.com/anza-xyz/agave/pull/14874) (merged 2026-08-28) makes `From<generated::Message> for VersionedMessage` check `config` before `versioned`, so V1 messages reconstruct as V1 with the transaction config intact.
* Released in `solana-storage-proto` 4.2.2 — I checked the published crate source; the fix and its round-trip tests are in it.
* Pulled in here by #871, shipped as `@triton-one/yellowstone-grpc` 7.0.1.

So the napi decode path no longer downgrades V1 to V0, and `transactionConfig` survives.

One loose end: the CHANGELOG entry for 2026-08-31 lists only `yellowstone-grpc-geyser 15.2.1`. The nodejs 7.0.1 release and this fix aren't recorded there, so anyone bumping the client has no way to see the V1 fix landed. Happy to open a one-line PR for it if that's useful.

Closing.

PR #868, opened 2026-08-26
Extract convert_from/convert_to into yellowstone-grpc-convert; drop the Node client's solana-storage-proto round-trip

serejke:
Implements #867.

## What

- **New `yellowstone-grpc-convert` crate** containing `convert_from` and `convert_to`, moved verbatim from `yellowstone-grpc-geyser/src/plugin/` (git detects the move; `convert_to.rs` is byte-identical). Clients can depend on it without pulling in the validator plugin crate.
- **`yellowstone-grpc-geyser` re-exports both modules at their old paths** (`plugin::convert_from` / `plugin::convert_to`), so the geyser crate's public API and `examples/rust` are unchanged. Geyser drops the five dependencies that were only used by the moved code (`solana-account`, `solana-account-decoder`, `solana-transaction-context`, `solana-transaction-error`, `wincode`).
- **napi uses the conversions directly.** `encode_tx` and `encode_deshred_tx` previously re-serialized the yellowstone proto message with `encode_to_vec()` and re-parsed the bytes as agave's `solana_storage_proto` twin type, using agave's `From` conversions. That inherited agave's V1 bug (#866: V1 transactions reported as V0 with `transactionConfig` dropped — the bug this repo already fixed on the geyser side in #855), relied on the two `solana-storage.proto` schemas staying wire-identical with nothing enforcing it, and paid a full protobuf re-serialize + re-parse per transaction. All three go away.
- **`solana-storage-proto` removed from the napi crate.** The geyser crate's `bench`/dev usages are untouched.
- **Fix: `create_tx_meta` now respects `inner_instructions_none` and `log_messages_none`** (`return_data_none` was already handled). Without this, the swap would have turned "not recorded" into empty arrays; it was also a latent bug for existing `convert_from` users — `examples/rust` rendered `[]` where the plugin signalled none. `encode_deshred_tx` relies on these flags for its synthetic meta.
- Tests: the `_none` flags in both states, and a V1 message round-tripping `convert_to::create_message` -> `convert_from::create_message` with its `TransactionConfig` intact.

## napi dependency delta (measured with `cargo tree -e normal`)

- removed: `solana-storage-proto v4.2.0`, `solana-serde v3.0.0`
- added: `yellowstone-grpc-convert v0.1.0` (path crate; all of its dependencies were already in the napi tree)

`solana-transaction-status` now needs its `agave-unstable-api` feature declared explicitly in the napi and convert crates — it was previously enabled transitively by `solana-storage-proto`.

## Verification (Rust 1.96.1 per `rust-toolchain.toml`)

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets`
- `cargo test --workspace` — 361 passed, 0 failed
- `cargo build -p yellowstone-grpc-napi` (standalone, so the binding builds without the geyser crate in the graph)
- `cargo check -p yellowstone-grpc-geyser --features bench --all-targets`


lvboudre:
Fixed in tyescript client v7.0.1

PR #878, opened 2026-09-07
feat: wire in blockfooter message

Ozodimgba:


lvboudre (commented):
Please add e2e test for block footer.

PR #880, opened 2026-09-09
fix(geyser): decrement message_queue_size on startup snapshot drop

### LINKED ISSUE #840: `message_queue_size` gauge leaks +1 per slot since v14.2.2 (BlockMeta partition drained without decrement)
## Summary

Since `v14.2.2+solana.4.1.0` (243d008), the `yellowstone_grpc_geyser_message_queue_size` gauge grows monotonically by exactly **+1 per slot (~2.5/s)** and never drains. The pipeline itself is healthy — this is a metrics accounting bug only, but it makes the gauge useless for backpressure monitoring (it fired our alerts as an apparent slow leak).

## Root cause

#838 moved `BlockMeta` handling so it is delegated to the block-reconstruction loop from inside the geyser loop (fixing the account-updates-during-freeze race). The `PartitionedBuffer` it introduced in `geyser_loop` diverts `Message::BlockMeta` into a separate `blockmeta_batch` slot instead of `message_batch`.

The gauge accounting was not updated to match:

- **Enqueue side**: `PluginInner::send_message` increments the gauge for *every* message sent to the geyser loop, including `BlockMeta` ([`plugin/entry.rs:58`](https://github.com/rpcpool/yellowstone-grpc/blob/243d008a1c176492c8cd48ee225a4f9502e1ba60/yellowstone-grpc-geyser/src/plugin/entry.rs#L58)).
- **Drain side**: only the regular batch is decremented — `metrics::message_queue_size_dec_by(buffer.message_batch.len() as i64)` ([`grpc.rs:1256`](https://github.com/rpcpool/yellowstone-grpc/blob/243d008a1c176492c8cd48ee225a4f9502e1ba60/yellowstone-grpc-geyser/src/grpc.rs#L1256)). The `blockmeta_batch.take()` branch ([`grpc.rs:1269`](https://github.com/rpcpool/yellowstone-grpc/blob/243d008a1c176492c8cd48ee225a4f9502e1ba60/yellowstone-grpc-geyser/src/grpc.rs#L1269)) forwards the message to block reconstruction without ever calling `message_queue_size_dec()`.

Net effect: the gauge leaks +1 per `BlockMeta`, i.e. +1 per slot, forever. (The one standalone `message_queue_size_dec()` at `grpc.rs:1681` belongs to the snapshot/startup channel and is unrelated.)

## Observed

Production RPC node running `v14.2.2+solana.4.1.0`:

- gauge grows at ~2.4/s, matching slot rate;
- reading ≈ plugin uptime × 2.5 exactly (68 min uptime → ~9.8k);
- meanwhile `block_reconstruction_queue_size` = 0 and subscriber queues are empty — messages are actually being consumed, node is 1–2 slots behind the cluster tip as usual.

## Affected versions

- `v14.2.0` / `v14.2.1`: **not affected** (single batch, uniform `dec_by`).
- `v14.2.2`: affected (first release with the partition).
- `master` (8e747a1): affected.

## Impact

Metric-only — the geyser channel is unbounded mpsc, so there is no real backpressure or memory growth from this. But the gauge no longer reflects real queue depth and looks exactly like a slow queue leak on dashboards.

## Fix

One line: decrement the gauge in the `blockmeta_batch.take()` branch, mirroring the accounting of the batch branch (decrement on drain, regardless of downstream send result). PR incoming.

Tyagiquamar:
Fixes #840

### DIFF
--- a/yellowstone-grpc-geyser/src/plugin/entry.rs
+++ b/yellowstone-grpc-geyser/src/plugin/entry.rs
@@ -301,7 +301,12 @@ impl GeyserPlugin for Plugin {
 
     fn notify_end_of_startup(&self) -> PluginResult<()> {
         self.with_inner(|inner| {
-            let _snapshot_channel = inner.snapshot_channel.lock().unwrap().take();
+            if let Some(snapshot_channel) = inner.snapshot_channel.lock().unwrap().take() {
+                let count = snapshot_channel.len();
+                if count > 0 {
+                    metrics::message_queue_size_dec_by(count as i64);
+                }
+            }
             Ok(())
         })
     }


Tyagiquamar:
Hi, friendly ping for review on this PR when you have a moment. Happy to address any feedback. Thanks!

Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

PR #886, opened 2026-09-21
feat(geyser): support entry and deshred parent update callbacks

Ozodimgba:


yellowstone-grpc-geyser/src/plugin/filter/message.rs
```diff
@@ -269,16 +273,17 @@ pub type FilteredUpdateFilters = SmallVec<[FilterName; 4]>;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum FilteredUpdateOneof {
-    Account(FilteredUpdateAccount),                     // 2
-    Slot(FilteredUpdateSlot),                           // 3
-    Transaction(FilteredUpdateTransaction),             // 4
+    EntryUpdateParent(Arc<MessageEntryUpdateParent>), // tag 13
```
Synestraa: although the order of tag vs enum was already out of place here, EntryUpdateParent doesn't belong at the top of the list here, could you put it under Entry or under blockfooter?

yellowstone-grpc-geyser/src/plugin/entry.rs
```diff
@@ -465,6 +467,33 @@ impl GeyserPlugin for Plugin {
...
 
+    fn notify_entry_update_parent(
+        &self,
+        update_parent: ReplicaEntryUpdateParentInfoVersions,
+    ) -> PluginResult<()> {
+        self.with_inner(|inner| {
+            let ReplicaEntryUpdateParentInfoVersions::V0_0_1(info) = update_parent;
+            let message =
+                Message::EntryUpdateParent(Arc::new(MessageEntryUpdateParent::from_geyser(info)));
+            inner.send_message(message);
```
Synestraa, line 478: block_machine.add will ignore this update, therefore subscribers wont see it for confirmed/finalized. should either modify block reconstruction to support this type of update, or add confirmed/finalized send here similar to slot updates. the update itself may be relevant to block reconstruction so I'm leaning towards that

Synestraa (changes requested):

yellowstone-grpc-geyser/src/plugin/entry.rs
Ozodimgba: i think I ship what i have for block reconstruction in a separate PR, so that i make a boundary between scope

Synestraa (approved):

